PR #5, opened 2021-02-25
Make pythonIoc pip installable

thomascobb:
At the moment it is an EPICS module that embeds Python. This PR will make it a Python module that embeds an EPICS IOC.


thomascobb:
@Araneidae this is working apart from a couple of things with devIocStats

1. We don't use the epics build system to make ioc.db, and dbLoadDatabase doesn't like ioc.template because it contains the msi "substitute" command. I can get round this in Python
2. Loading devIocStats gives: `Error (511,35) PV: TS-DI-TEST-01:SYSRESET Proc subroutine (SNAM)`. This is a mystery to me as the relevant routine appears when you do `strings` on the extension that I built. Any ideas?

thomascobb:
I'll add the CI and more tests when these things are working...

thomascobb:
@mdavidsaver notes that 2. is because registerAllRecordDeviceDrivers() isn't handling function(). He'll look at modifying this in EPICS base...

softioc/extension.c
```diff
@@ -85,33 +72,36 @@ void get_field_offsets(
...
-                field_offset[i] = dbentry.pflddes->offset;
-                field_size[i]   = dbentry.pflddes->size;
-                field_type[i]   = dbentry.pflddes->field_type;
-            }
-        }
+        PyObject *ost = Py_BuildValue("iii",
+            dbentry.pflddes->offset,
+            dbentry.pflddes->size,
+            dbentry.pflddes->field_type);
+        PyDict_SetItemString(dict, field_name, ost);
```
mdavidsaver, line 79: I think a reference to `ost` is being leaked here.  From https://docs.python.org/3/c-api/dict.html#c.PyDict_SetItemString

>  This function does not steal a reference to val.

Araneidae: Nice catch, I'm sure you're right here.

softioc/extension.c
```diff
@@ -77,6 +80,7 @@ static PyObject *get_field_offsets(PyObject *self, PyObject *args)
             dbentry.pflddes->size,
             dbentry.pflddes->field_type);
         PyDict_SetItemString(dict, field_name, ost);
+        Py_XDECREF(ost);
```
thomascobb, line 83: Is this the right fix?

softioc/extension.c
```diff
@@ -20,14 +20,17 @@
 
 /* Helper for function below. */
 #define ADD_ENUM(dict, name) \
-    PyDict_SetItemString(dict, #name, PyInt_FromLong(name))
+    tmp = PyInt_FromLong(name); \
+    PyDict_SetItemString(dict, #name, tmp); \
+    Py_XDECREF(tmp);
```
thomascobb, line 25: And do we need this here too?

softioc/extension.c
Araneidae: Yes, looks right to me ... except, you can use `Py_DECREF` instead of `Py_XDECREF` -- you _know_ that `ost` is not null at this point.

softioc/extension.c
Araneidae: Ick.  Yes (and you can use `Py_DECREF` again).

To reduce the ick factor at this point I might just make a helper function
```c
static void add_enum(PyObject *dict, const char *name, long value)
{
    PyObject *py_value = PyInt_FromLong(value);
    PyDict_SetItemString(dict, name, py_value);
    Py_DECREF(py_value);
}
```
and either call that directly or simplify `ADD_ENUM` to
```c
#define ADD_ENUM(dict, name) add_enum(dict, #name, name)
```

An interesting thought on terminology.  The documentation for `PyDict_SetItemString` says
> This function _does not_ steal a reference to `val`.

I guess today I'd rather reword that as
> This function does not take ownership of `val`

softioc/extension.c
```diff
@@ -18,9 +18,16 @@
 #define PyInt_FromLong(value)   PyLong_FromLong(value)
 #endif
 
+/* Reference stealing version of PyDict_SetItemString */
+static void set_dict_item_steal(PyObject *dict, const char *name, PyObject *py_value)
```
thomascobb, line 22: Any better?

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "wheel", "setuptools_dso", "epicscorelibs>=7.0.4.99.0.0a1"]
+build-backend = "setuptools.build_meta"
```
Araneidae, line 3: Missing eol at end of file

python/softioc/pvlog.py
```diff
@@ -1,10 +0,0 @@
-# Support for PV put logging
-#
-# Importing this module configures PV put logging
-
-import os
-from . import imports
-
-imports.asSetFilename(
-    os.path.join(os.path.dirname(__file__), '..', 'access.acf'))
-imports.asTrapWriteRegisterListener(imports.EpicsPvPutHook)
```
Araneidae, line 10: Where's this functionality gone to?

softioc/extension.c
Araneidae: Guess so!

softioc/extension.c
```diff
@@ -85,33 +79,36 @@ void get_field_offsets(
...
+    long length;
+    if (!PyArg_ParseTuple(args, "shnl", &name, &dbrType, &pbuffer, &length))
+        return NULL;
+
     struct dbAddr dbAddr;
     int rc = dbNameToAddr(name, &dbAddr);
     if (rc == 0)
         rc = dbPutField(&dbAddr, dbrType, pbuffer, length);
-    return rc;
+    return Py_BuildValue("i", rc);
```
Araneidae, line 111: Interesting issue here.  Turns out we ignore the return code at the point of call:
https://github.com/dls-controls/pythonIoc/blob/31fd6117d1d57f90273f75c77c16a2db67f63023/softioc/device.py#L205-L206
I guess we have two choices:
1. raise an exception instead of returning an ignored status code
2. Ignore the failure altogether

The calling function https://github.com/dls-controls/pythonIoc/blob/31fd6117d1d57f90273f75c77c16a2db67f63023/softioc/device.py#L194 is part of the user API, and so it would probably be an improvement to tell the caller that they've made a mistake, so I favour turning this `Py_BuildValue` return into setting an exception and returning `NULL` on failure.

softioc/extension.c
```diff
@@ -205,3 +198,59 @@ void EpicsPvPutHook(struct asTrapWriteMessage *pmessage, int after)
         /* Just save the old value for logging after. */
         pmessage->userPvt = value;
 }
+
+static PyObject *install_pv_logging(PyObject *self, PyObject *args)
```
Araneidae, line 202: Oh, this is where pv logging has gone.

Don't forget, still need to be able to initialise and enable PV logging by simply doing `import softioc.pvlog`.

softioc/extension.c
```diff
@@ -205,3 +198,59 @@ void EpicsPvPutHook(struct asTrapWriteMessage *pmessage, int after)
...
+        PyObject *mod = Py_InitModule("softioc._extension", softioc_methods);
+#endif
+        if(mod) {
+        }
+#if PY_MAJOR_VERSION >= 3
+    return mod;
+#else
+    (void)mod;
+#endif
+}
```
Araneidae, line 256: No EOL at end of file!

softioc/pvlog.py
```diff
@@ -0,0 +1,9 @@
+# Support for PV put logging
+#
+# Importing this module configures PV put logging
+
+import os
+from . import imports
+
+imports.install_pv_logging(
+    os.path.join(os.path.dirname(__file__), 'access.acf'))
```
Araneidae, line 9: Oh goodness, sorry about all that.  Here it is.

It didn't appear as a rename in the earlier part of the commit.

softioc/extension.c
thomascobb: This now raises an exception
```
>>> t_ao.set(b"bad value")
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/tom/Programming/pythonIoc/softioc/device.py", line 206, in set
    db_put_field(
  File "/home/tom/Programming/pythonIoc/softioc/imports.py", line 24, in db_put_field
    return _extension.db_put_field(name, dbr_type, pbuffer, length)
RuntimeError: dbPutField failed for TS-DI-TEST-01:AO
```

Araneidae:
Is this the same build commit over and over again?  I've been getting the notifications since last week!  Guess you're chasing the missing CI failures?

thomascobb:
Yes, hundreds of modifications to the build step, but each needs to be pushed to trigger CI. I'm getting there though, just the windows build left to do...

thomascobb:
Ok, this should be ready to review again, just docs left to do

softioc/device.py
```diff
@@ -1,17 +1,33 @@
...
+    # processing doesn't interfere with other callback processing.
+    global dispatcher_callback
+    dispatcher_callback = cothread.cothread._Callback()
+
+
+def use_asyncio():
+    from .asyncio_callback import AsyncioCallback
+    global dispatcher_callback
+    dispatcher_callback = AsyncioCallback()
+    dispatcher_callback.start()
```
thomascobb, line 28: `assert dispatcher_callback is None`

softioc/device.py
```diff
@@ -1,17 +1,33 @@
...
 from . import alarm
 from .fields import DbfCodeToNumpy, DbrToDbfCode
 from .imports import dbLoadDatabase, recGblResetAlarms, db_put_field
 from .device_core import DeviceSupportCore, RecordLookup
 
 
+dispatcher_callback = None
+
+
+def use_cothread():
```
thomascobb, line 16: Delete these and move the logiv to `softIoc(dispatcher=None)`

thomascobb:
Once we have a 1.4.1 release of epicsdbbuilder, then I can regenerate Pipfile.lock with DLS deps, then I think I'm done here

Araneidae (commented):
Forgot about this, don't know if these are relevant anymore!

softioc/extension.c
```diff
@@ -205,3 +200,60 @@ void EpicsPvPutHook(struct asTrapWriteMessage *pmessage, int after)
...
+#else
+        PyObject *mod = Py_InitModule("softioc._extension", softioc_methods);
+#endif
+        if(mod) {
+        }
+#if PY_MAJOR_VERSION >= 3
+    return mod;
+#else
+    (void)mod;
+#endif
```
Araneidae, line 258: This is very strange looking code!  The indentation is messed up, there's an empty statement in the middle, and we have a repeated `#if` test.  I'd suggest rewriting this as:
```c
#if PY_MAJOR_VERSION >= 3
    return PyModule_Create(&softioc_module);
#else
    Py_InitModule("softioc._extension", softioc_methods);
#endif
```

PR #14, opened 2021-07-01
Docs re-write, including for asyncio support

AlexanderWells-diamond:
Wrote several how-to guides, re-wrote the "Creating an IOC" tutorial, wrote example code for IOCs using cothread and asyncio, and another example for how to remotely read from those example IOCs.

### DIFF
--- a/docs/how-to/make-publishable-ioc.rst
+++ b/docs/how-to/make-publishable-ioc.rst
@@ -1,7 +1,7 @@
 Create a Publishable IOC
 ========================
 
-As seen in :doc:`../tutorials/creating-an-ioc`, a single Python script can be an IOC.
+As seen in `../tutorials/creating-an-ioc`, a single Python script can be an IOC.
 It is also possible (and the most common situation) to have an entire Python module
 comprising an IOC. This guide explains both, as well as how to publish an IOC within
 the DLS environment.
--- a/docs/how-to/make-publishable-ioc.rst
+++ b/docs/how-to/make-publishable-ioc.rst
@@ -1,0 +1,65 @@
+Create a Publishable IOC
+--------------------------
+
+As the example script above shows, a single Python script can be an IOC.
+However, to fit into the DLS framework for publishing IOCs in ``/dls_sw/prod`` a
+bit more structure is needed.  I recommend at least four files as shown:
+
+``Makefile``
+    This file is necessary in order to run ``dls-release.py``, and needs to have
+    both ``install`` and ``clean`` targets, but doesn't need to actually do
+    anything.  Thus the following content for this file is enough::
+
+        install:
+        clean:
+
+``start-ioc``
+    An executable file for starting the IOC needs to be created.  I recommend
+    that this consist of the following boilerplate::
+
+        #!/bin/sh
+
+        PYIOC_VER=2-6
+        EPICS_VER=3.14.12.3
+
+        PYIOC=/dls_sw/prod/R$EPICS_VER/support/pythonSoftIoc/$PYIOC_VER/pythonIoc
+
+        exec $PYIOC ioc_entry.py "$@"
+
+    Here I have given the startup script for the IOC the name ``ioc_entry.py``.
+    This name should be replaced by any appropriate name.
+
+``ioc_entry.py``
+    I recommend that the top level Python script used to launch the IOC contain
+    only ``pkg_resources.require`` statements, simple code to start the body
+    of the IOC, and it should end with standard code to start the IOC.  The
+    following structure can be followed (here I've assumed that the rest of the
+    IOC is in a single file called ``ioc_body.py``::
+
+        from pkg_resources import require
+
+        require('cothread==2.12')
+        require('epicsdbbuilder==1.0')
+        # Any other requires needed by this IOC
+
+        from softioc import softioc
+
+        # Do whatever makes sense to create all the PVs and get ready to go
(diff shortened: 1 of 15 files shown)


docs/how-to/make-publishable-ioc.rst
```diff
@@ -0,0 +1,35 @@
+Create a Publishable IOC
+========================
+
+As seen in :doc:`../tutorials/creating-an-ioc`, a single Python script can be an IOC.
```
thomascobb, line 4:
```suggestion
As seen in `creating-an-ioc`, a single Python script can be an IOC.
```
Does this work?

docs/how-to/read-data-from-ioc.rst
```diff
@@ -0,0 +1,31 @@
...
+======================
+
+This guide explains how to read data from an IOC in a separate Python program. 
+
+.. note:: 
+    Please ensure your firewall allows both TCP and UDP traffic on ports 5064 and 5065. 
+    These are used by EPICS for channel access to the PVs.
+
+
+To start, run the :mod:`cothread` IOC from :doc:`../tutorials/creating-an-ioc` or the
```
thomascobb, line 11:
```suggestion
To start, run the `cothread` IOC from `creating-an-ioc` or the 
```
etc etc

docs/how-to/read-data-from-ioc.rst
```diff
@@ -0,0 +1,31 @@
...
+interactive shell.
+
+We will read data from that IOC using this script:
+
+.. literalinclude:: ../examples/example_read_from_ioc.py
+
+.. note::
+    You may see warnings regarding the missing "caRepeater" program. This is an EPICS tool
+    that is used to track when PVs start and stop. It is not required for this simple example,
+    and so the warning can be ignored.
```
thomascobb, line 22: According to https://github.com/dls-controls/aioca/pull/12#issuecomment-868862584 then if you upgrade the epicscorelibs dependency to `epicscorelibs>=7.0.5.99.1.0a2` (in pyproject.toml) then remove the Pipfile.lock and run `pipenv --rm && pipenv install --dev`` then these warnings should go away. Please could you try that?

docs/tutorials/creating-an-ioc.rst
```diff
@@ -233,17 +133,5 @@ reading and writing the current value of the record.  For IN records calling
...
-which must be called in this order.  After calling
-:func:`~softioc.builder.LoadDatabase` it is no longer possible to create PVs.
-
-It is sensible to start any server background activity after the IOC has been
-initialised by calling :func:`~softioc.softioc.iocInit`.  After this has been
-done (:class:`cothread.Spawn` is recommended for initiating persistent background
-activity) the top level script must pause, as as soon as it exits the IOC will
-exit.  Calling :func:`~softioc.softioc.interactive_ioc` is recommended for this
-as the last statement in the top level script.
+..  _cothread: https://github.com/dls-controls/cothread
```
thomascobb, line 136: You shouldn't need these links as intersphinx should pick up `cothread` and `epicsdbbuilder`

thomascobb (changes requested):
Looking good, hopefully you can remove lots of `:role:` prefixes, but the content looks fine

docs/how-to/make-publishable-ioc.rst
AlexanderWells-diamond: I can't remove the "../tutorials/" part without encountering this error:
`WARNING: 'any' reference target not found: creating-an-ioc`

From my understanding of the [:any: ](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-any)and the [:doc: ](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-doc) roles, it looks like the full path is always required but the :doc: can be inferred from a lack of any role prefix. 

I'll go through and delete all the role tags I can.

docs/how-to/read-data-from-ioc.rst
AlexanderWells-diamond: Doing this update causes tests to fail due to a missing import. I also see the same error when just trying to start up my example IOCs:
`  File "docs/examples/example_cothread_ioc.py", line 2, in <module>
    from softioc import softioc, builder
ModuleNotFoundError: No module named 'softioc'`

I can verify that this is entirely caused by changing the epicscorelibs dependency - swapping the required version back and reinstalling the pipenv makes the tests (and my example IOCs) work again.
I can't say I understand this error at the moment, so I'll investigate it a little more.

docs/tutorials/creating-an-ioc.rst
AlexanderWells-diamond: Resolved in commit a873d0e

docs/how-to/read-data-from-ioc.rst
AlexanderWells-diamond: Resolved in commit a873d0e

docs/how-to/make-publishable-ioc.rst
AlexanderWells-diamond: Resolved in commit a873d0e

docs/how-to/read-data-from-ioc.rst
thomascobb: Oops, you need to do `dls-py3 download-one-dependency epicscorelibs 7.0.5.99.1.0a2` first

PR #20, opened 2021-07-29
Documentation for iocbuilder record creation

AlexanderWells-diamond:
Adding documentation for using IOC builder records, mixing and matching with pythonSoftIOC records, and creating a calc field and result.


AlexanderWells-diamond:
Note that I won't merge this for the time being as the CI seems to be completely broken. It's also broken on the master branch too, so its likely something to do with the machines we're running it on.

For the record, the last run of both Docs CI and Code CI on the master branch did pass, but I pressed the "rerun all jobs" button on the Code CI run without realising that it would delete that run and completely replace it.

AlexanderWells-diamond:
I think I've now fixed the CI - it seems some unpinned dependencies had breaking updates and so required pinning to continue working. I don't entirely understand the setuptools_dso pin though, as I can't find what's actually changed to break it. It just seems that broken builds were picking up `setuptools-dso-2.1`, whereas when we use `2.0` it works.

thomascobb:
> I think I've now fixed the CI - it seems some unpinned dependencies had breaking updates and so required pinning to continue working. I don't entirely understand the setuptools_dso pin though, as I can't find what's actually changed to break it. It just seems that broken builds were picking up `setuptools-dso-2.1`, whereas when we use `2.0` it works.

I added some code that makes auditwheel work here: https://github.com/dls-controls/pythonSoftIOC/commit/416b66b22cc4d74cac5b27d65693c2f4cb063b01

It looks like @mdavidsaver added the same things in setuptools-dso 2.1. The correct thing to do is to use setuptools-dso >= 2.1, then delete my stuff. Should still work.

@mdavidsaver do you have an example that uses auditwheel and setuptools-dso 2.1 we can copy?

thomascobb:
So, revert `setup.py` from that commit, then add `CIBW_ENVIRONMENT_LINUX: SETUPTOOLS_DSO_PLAT_NAME=manylinux1_x86_64` in `code.yml` as per https://cibuildwheel.readthedocs.io/en/stable/options/#environment and https://mdavidsaver.github.io/setuptools_dso/usage.html?highlight=auditwheel#using-with-manylinux

mdavidsaver:
> ... It just seems that broken builds were picking up `setuptools-dso-2.1`, whereas when we use `2.0` it works.

This sounds like a bug with setuptools-dso.  My goal is that minor version increments should remain fully compatible.  So eg. a reasonable version restriction now should be `setuptools-dso<=3.0a1`.  Can you elaborate on the breakage and/or open a [ticket](https://github.com/mdavidsaver/setuptools_dso/issues)?

> @mdavidsaver do you have an example that uses auditwheel and setuptools-dso 2.1 we can copy?

I don't use auditwheel, and I don't think it is usable with setuptools-dso.  Happily, as long as `$SETUPTOOLS_DSO_PLAT_NAME` continues to work, I don't think auditwheel is necessary.

thomascobb:
> Can you elaborate on the breakage
@mdavidsaver this hack worked in setuptools-dso 2.0:
https://github.com/dls-controls/pythonSoftIOC/blob/a9d72019dd05838cfa84f4161fe008d81af8bc98/setup.py#L103

I believe you put this same functionality into setuptools-dso 2.1, so we no longer need it. Leaving the hack in place causes this:
https://github.com/dls-controls/pythonSoftIOC/runs/3192523814

It may or may not be a bug, but I don't think anyone else will come across it, so I'm happy to just update to setuptools-dso 2.1

> I don't use auditwheel, and I don't think it is usable with setuptools-dso. 
Oops, I meant cibuildwheel. Do you have an example that uses cibuildwheel and setuptools-dso 2.1 we can copy?

mdavidsaver:
> AssertionError: cmdclass[<class '__main__.Wheel'>] must be subclass of <class 'setuptools_dso.dsocmd.bdist_wheel'>

Ah, I see.  It broke a downstream build unnecessarily, so imo. this is as a bug.  I should have made sure you saw https://github.com/mdavidsaver/setuptools_dso/pull/18.  I think this check should be "demoted" to a warning.


> Oops, I meant cibuildwheel. Do you have an example that uses cibuildwheel and setuptools-dso 2.1 we can copy?

Still no.  I haven't revisited it recently.  I'll be interested to see if/when you have something working :)

AlexanderWells-diamond:
It looks like my last commit passes CI successfully, so this is should be ready for (re-) review @thomascobb .

PR #22, opened 2021-08-09
Allow waveform.set() to be called before iocInit()

### LINKED ISSUE #21: Unable to run waveform.set() before softioc.iocInit()
In a patch from 2019 (CGP-182, 944e0725), a change was made to the waveform 'set' method. As part of this change, it now uses the 'dtype' attribute.

This attribute is only set when the 'init_record' method is called, which is triggered by the softioc.iocInit() call for IOC startup.

Here is my particular use-case:

The module I am working on has its behaviour split into a number of separate 'servers', each with their own EPICS database setup and associated logic. The 'softioc.iocInit()' method is only called by the top-level program, after each of these servers has had a chance to run its own database setup.

The unit tests do not call 'softioc.iocInit()'. They can, for example, allow the servers to run waveform.set() and then check the parameters afterwards.

This has problems after the above bug-fix, as dtype is not set. It seems sensible to me that unit tests should not need to have a running IOC in order to operate. What is your opinion on allowing get() and set() functions to be called before iocInit()? Do you think the above is a good model for unit testing? If it's not, what alternatives should be used?

MJGaughran:
Fixes #21.

With CGP-182 (944e072), waveform.set() was modified to use the dtype attribute, which is only created after softioc.iocInit() is called.

This change adds a default for dtype, preventing this issue.

### DIFF
--- a/softioc/device.py
+++ b/softioc/device.py
@@ -293,6 +293,8 @@ class WaveformBase(ProcessDeviceSupportCore):
     #   NELM    Length of allocated array in number of elements
     #   NORD    Currently reported length of array (0 <= NORD <= NELM)
     _fields_ = ['UDF', 'FTVL', 'BPTR', 'NELM', 'NORD']
+    # Allow set() to be called before init_record:
+    dtype = None
 
     def init_record(self, record):
         self.dtype = DbfCodeToNumpy[record.FTVL]
--- a/tests/sim_records.py
+++ b/tests/sim_records.py
@@ -42,6 +42,8 @@ def update_sin_wf(value):
     sin_wf.set(numpy.sin(
         numpy.linspace(0, 2*numpy.pi*sin_ph.get(), sin_len.get())))
 sin_wf = Waveform('SIN', datatype = float, length = 1024)
+# Check we can update its value before iocInit as per #22
+sin_wf.set([1, 2, 3])
 sin_len = longOut(
     'SINN', 0, 1024, initial_value=1024, on_update=update_sin_wf)
 sin_ph = aOut('SINP', initial_value = 0.0, on_update = update_sin_wf)


softioc/device.py
```diff
@@ -293,6 +293,7 @@ class WaveformBase(ProcessDeviceSupportCore):
     #   NELM    Length of allocated array in number of elements
     #   NORD    Currently reported length of array (0 <= NORD <= NELM)
     _fields_ = ['UDF', 'FTVL', 'BPTR', 'NELM', 'NORD']
+    dtype = None
```
Araneidae, line 296: Probably worth adding a one line comment above this to explain why!

(Sorry, didn't mention this before; best to force push the comment, please.)

Araneidae:
@thomascobb , didn't realise our CI was so broken!  I asked @MJGaughran to make this a pull request so that CI would give it a sanity check ... but that seems to be a bit busted 😞

softioc/device.py
MJGaughran: Just to prevent lots of forced pushes, is this okay:
```
_fields_ = ...
# Allow set() to be called before init_record():
dtype = None
```

Araneidae:
Am happy with this, but will wait for CI to be fixed I guess...

AlexanderWells-diamond:
PR #20 contains both documentation fixes and fixes for the CI. Once that's merged this should clear up. It's also pending review at the moment.

thomascobb:
I'll make a couple of docs tweaks to #20 then merge it

softioc/device.py
MJGaughran: I've made the changes as I described above. Easy enough to modify if you're not happy with the comment

Araneidae: That's odd... I think the comment line just vanished!

MJGaughran: Okay, let's just pretend that never happened. I think I've fixed it now...

thomascobb:
I added a test too and merged in master

PR #28, opened 2021-08-12
Changes required for integration into tickit simulation framework

thomascobb:
### Make sure we use epicscorelibs libca in cothread
Move this as early as we can to avoid loading 2 versions of libca by mistake

### Allow AsyncioDispatcher to take existing event loop
Need this if we want to embed IOC in existing asyncio app

### Add ability to add arbitrary lines to Db file and test 
Need this so we can read arbitrary db file, remove DTYP lines, add pythonSoftIOC records for logic of those records, then have it appear in the db file that is loaded in the IOC
See [tests/sim_asyncio_ioc_override.py](https://github.com/dls-controls/pythonSoftIOC/pull/28/files#diff-57b1ece2045fef10c0ac883d66e9302f13b2e069b0a25763896f1dd2a6c27ba0) for an example of this


thomascobb:
@Araneidae Mac builds (apart from python 2.7) seem to be stuck in a queue somewhere on Github Actions, but I think they will work fine, so this is ready for review

Araneidae (commented):
I realise I have no idea what this patch is doing, certainly no idea what "tickit" is...  Not sure I can contribute very much

softioc/__init__.py
```diff
@@ -5,6 +5,10 @@
 from epicscorelibs.ioc import \
     iocshRegisterCommon, registerRecordDeviceDriver, pdbbase
 
+# Do this as early as possible, in case we happen to use cothread
+# as it's harmless if we don't
+import epicscorelibs.path.cothread  # noqa
```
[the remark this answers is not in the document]
thomascobb: All this does is declare a single environment variable, so is harmless if `cothread` isn't installed

softioc/softioc.py
```diff
@@ -240,10 +240,9 @@ def __call__(self):
 exit = Exiter()
 command_names.append('exit')
 
+# For backwards compatibility
+dbLoadDatabase = builder.dbLoadDatabase
```
thomascobb, line 244: @Araneidae do you think I need to have this line in case anyone is calling `softioc.dbLoadDatabase`?

tests/sim_asyncio_ioc_override.py
```diff
@@ -0,0 +1,33 @@
...
+
+    # Override DTYPE and OUT, and provide a callback
+    gain = builder.boolOut("GAIN", on_update=print)
+    softioc.devIocStats(parsed_args.prefix)
+
+    # Run the IOC
+    builder.LoadDatabase()
+    event_loop = asyncio.get_event_loop()
+    softioc.iocInit(asyncio_dispatcher.AsyncioDispatcher(event_loop))
+    event_loop.run_forever()
```
thomascobb, line 33: @Araneidae I wanted to see if you thought this was the best approach to solving the "load existing database, strip out DTYP, then load some pythonSoftIOC override records on top"

softioc/builder.py
```diff
@@ -181,6 +185,14 @@ def WaveformOut(name, *value, **fields):
 
 _DatabaseWritten = False
 
+def AddDatabaseLine(line, macros={}):
```
thomascobb, line 188: This lets you add arbitrary lines to the database file. Is it named correctly, and could we do the macro replacement in a more efficient way?

softioc/__init__.py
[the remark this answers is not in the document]
Araneidae: But what is `epicscorelibs.path`?

I ended up navigating to [here](https://github.com/mdavidsaver/epicscorelibs/blob/master/src/python/epicscorelibs/path/cothread.py) to answer this question.  Ah well, a bit cryptic, but now I see.

thomascobb (commented):
I've commented on the lines I'd like reviewed

softioc/__init__.py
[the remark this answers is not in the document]
thomascobb: Maybe I need to put a better comment in
```suggestion
# This will set the CATOOLS_LIBCA_PATH environment variable in case we use cothread.catools
import epicscorelibs.path.cothread  # noqa
```

softioc/softioc.py
Araneidae: Well, it is in the documented api, so yes, I think it would be better to leave it in.  After all, this is the method I was expecting you to be using...

tests/sim_asyncio_ioc_override.py
Araneidae: Not a huge fan.  Building a database up line by line is very error prone, there's nothing stopping you from creating a horribly malformed file this way.

This is, essentially a hack.  As such I'd suggesting being a bit more honest about the hack and not add half-baked methods to the API.  In this particular case what it looks like you want to do is:

- Process `hw_records.db` to remove all lines containing any `field DTYP` definitions.  Put the result in a temporary file (just like `LoadDatabase` does under the hood...).
- Use `dbLoadDatabase` to load the hacked file and apply any MSI style substitutions.

softioc/builder.py
Araneidae: As I've commented above, I think I hate this API.  If you're going to build a database incrementally, you need to do it in a structured way.  Maybe parse the .db file and use the iocbuilder to rebuild it?  Not nice, I agree...

tests/sim_asyncio_ioc_override.py
thomascobb: This sounds like a better solution, I'll try it out

tests/sim_asyncio_ioc_override.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Load the base records without DTYP fields
+    with open(Path(__file__).parent / "hw_records.db", "rb") as inp:
+        with NamedTemporaryFile(suffix='.db', delete=False) as out:
+            for line in inp.readlines():
+                if not re.match(rb"\s*field\s*\(\s*DTYP", line):
+                    out.write(line)
+    softioc.dbLoadDatabase(
+        out.name, substitutions=f"device={parsed_args.prefix}")
+    os.unlink(out.name)
```
thomascobb, line 27: Any better?

Araneidae: Didn't think the name `out` would survive the scope of `with`, but yes, looks simple and to the point to me.

PR #34, opened 2021-08-31
Support alarm.*_ALARM in mbb{In,Out}

thomascobb:


tests/expected_records.db
```diff
@@ -1,7 +1,7 @@
 # This file was automatically generated on Fri 13 Aug 2021 16:49:03 BST.
-# 
+#
 # *** Please do not edit this file: edit the source file instead. ***
-# 
+#
```
Araneidae, line 4: Hmmm.  Something fishy about line 3 in this context!

thomascobb: Good point, will remove it

PR #38, opened 2021-09-15
Only use the calculated length if NELM was NOT provided during record creation

### LINKED ISSUE #37: Waveform records ignore NELM keyword if initial_value is set
If you specify both an `initial_value=...` and `NELM=...` keywords when creating a WaveformOut record, the `NELM` keyword is ignored and the waveform is created with a maximum length equal to the length of the data. 

I think the offending line is L169 of builder.py, which seems to override the passed in NELM without checking if it is set:
`fields['NELM'] = length`

There is a workaround: instead of `initial_value` and `NELM`, use something like this (assuming `data` is a numpy array)
```
length=required_length,
datatype=data.dtype,
```

And later do `myrecord.set(data)`

AlexanderWells-diamond:
Closes #37 

This PR is what I think the minimum fix is for the linked issue - it allows setting both `initial_value` and `NELM` during waveform record creation such that both keywords are respected.

It does not currently attempt to do anything regarding the "duplicate" `length` vs `NELM` or `value` vs `initial_value` keyword arguments. Using "mixed pairs" of Python and EPICS keywords will not work in all cases. Please let me know what, if anything, I should do with that.

A single test is written to cover the one case I originally identified - depending on the outcome of discussion I can write more, as appropriate.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "python.testing.pytestArgs": [],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
+}
--- a/docs/reference/api.rst
+++ b/docs/reference/api.rst
@@ -298,9 +298,10 @@ All functions return a wrapped `ProcessDeviceSupportIn` or
 
     If ``value`` is specified or if an `initial_value` is specified (only one of
     these can be used) the value is used to initialise the waveform and to
-    determine its field type and length.  If no initial value is specified then
-    the keyword argument ``length`` must be used to specify the length of the
-    waveform.
+    determine its field type and length (the inferred values may be overridden using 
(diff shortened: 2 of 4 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
thomascobb, line 1: I haven't committed any vscode config yet as @Araneidae doesn't use vscode. However as there are now two of us doing this, maybe we should. What do you think @Araneidae ?

AlexanderWells-diamond: Thinking about it, if we do go down that route we should probably just bring across all the .vscode settings from the template module, rather than what I've done here and just a small selection to get one particular piece working as I wanted.

Araneidae:
This is now obsoleted by PR #60

PR #60, opened 2021-11-22
Draft: Standardise record value setting/getting

### LINKED ISSUE #40: Unable to set StringOut record values
I seem to be unable to set the value of a StringOut record programmatically by calling `.set()`. This program demonstrates the issue:

```
# Import the basic framework components.
from softioc import softioc, builder, asyncio_dispatcher

# Create an asyncio dispatcher, the event loop is now running
dispatcher = asyncio_dispatcher.AsyncioDispatcher()

# Set the record prefix
builder.SetDeviceName("MY-DEVICE-PREFIX")

# Create some records
strout = builder.stringOut("FLIBBLE", initial_value="ABC")

# Boilerplate get the IOC started
builder.LoadDatabase()
softioc.iocInit(dispatcher)

strout.set("DEF")

# Finally leave the IOC running with an interactive shell.
softioc.interactive_ioc(globals())
```

The `strout.set()` line generates this exception:

```
 File "/home/eyh46967/dev/pythonSoftIoc/docs/examples/trial_stringout.py", line 17, in <module>
    strout.set("DEF")
  File "/home/eyh46967/dev/pythonSoftIoc/softioc/device.py", line 212, in set
    datatype, length, data, array = self.value_to_dbr(value)
  File "/home/eyh46967/dev/pythonSoftIoc/softioc/device.py", line 186, in value_to_dbr
    dbrtype = self.NumpyCharCodeToDbr[value.dtype.char]
KeyError: 'U'
```

The `device.py` code appears to be trying to convert the string `value` into a numpy array on line 173:
```
value = numpy.require(value, requirements = 'C')
```

Examining this array gives a `value.dtype.char` of `U`, which is unknown to the `NumpyCharCodeToDbr` mapping dictionary.

AlexanderWells-diamond:
This PR aims to ensure that the value returned from calling `.get()` on a record is always of the same form, regardless of whether the record has been initialised or not. 

Closes #40
TODO: This probably closes other issues too.


AlexanderWells-diamond:
PR parked pending removal of Python2 support:
https://github.com/dls-controls/pythonSoftIOC/issues/61

This will mean we won't need to handle Unicode in Python2.

AlexanderWells-diamond:
@Araneidae @thomascobb This PR is now ready for review.

softioc/device.py
```diff
@@ -219,14 +233,23 @@ def get(self):
...
 
 
-def _Device(Base, record_type, mlst=False, default=0, convert=True):
+def _Device(
+        Base,
+        record_type,
+        mlst=False,
+        default=0,
+        convert=True,
+        value_to_epics=None):
```
Araneidae, line 242: Looking at how the instances of this class are coming out, I'd be tempted to make `convert` and `value_to_epics` positional arguments without default (and let's rename `convert` to `epics_rc` and pass either `EPICS_OK` or `NO_CONVERT`).

Araneidae (commented):
Most of my comments so far are layout related.

I'm getting a bit distracted by inconsistent layout and use of spaces around `=`.  I'm guessing that spaces in function definitions is a lost call, but you'll find that all the preexisting code always uses spaces in all other contexts, particularly in function calls.

Secondly, please use 4 character indent except in those incredibly rare cases where it comes out to be particularly ugly.  You can always use a doc-string to make function argument indentation to look sensible.

I'd like the record device definitions to look more uniform and be easier to read, possibly even drop all(?) defaults?  Anyhow, try and get a layout that's easier on the eye here.

There's plenty of functionality for me to look at ... I'll get to it ...

softioc/builder.py
```diff
@@ -149,7 +149,15 @@ def _waveform(value, fields):
         # If a value is specified it should be the *only* non keyword
         # argument.
         value, = value
-        value = numpy.array(value)
+
+        if isinstance(value, str):
+            value = value.encode(errors='replace')
+
+        if isinstance(value, bytes):
```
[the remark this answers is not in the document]
AlexanderWells-diamond: Explicitly not - I'm converting any incoming string into bytes, and then that bytes into an array

softioc/device.py
```diff
@@ -236,18 +259,42 @@ class GenericDevice(Base):
...
+def _Device_Out(record_type, convert=True, mlst=True, **kargs):
+    return _Device(_Out, record_type, convert=convert, mlst=mlst, **kargs)
+
+
+def convert_to_int(value):
+    """Attempt to convert any type into its integer representation"""
+    if type(value) == c_int:
+        return value.value
+    # Can't just check value, as default value of 0 is false!
+    return int(value) if value is not None else None
```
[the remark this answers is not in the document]
AlexanderWells-diamond: After discussion, we decided to drop c_types and None from allowed inputs. These both now raise exceptions if attempted. See commit b615f6b

softioc/device.py
```diff
@@ -236,18 +259,42 @@ class GenericDevice(Base):
...
+
+def convert_to_float(value):
+    """Attempt to convert any type into its float representation"""
+    if type(value) == c_float:
+        return value.value
+    # Can't just check value, as default value of 0.0 is false!
+    return float(value) if value is not None else None
```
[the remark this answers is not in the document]
AlexanderWells-diamond: Resolved above.

softioc/device.py
```diff
@@ -236,18 +259,42 @@ class GenericDevice(Base):
...
+
+def truncate_string(value):
+    """Trim a string to EPICS 40 (39 with null byte) character limit"""
+    if isinstance(value, bytes):
+        value = value.decode(errors="replace")
+    return value[:39] if isinstance(value, str) else None
```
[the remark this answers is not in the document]
AlexanderWells-diamond: Resolved above.

softioc/device.py
AlexanderWells-diamond: `value_to_epics` is now a positional argument (as of b615f6b). Do you still want the changes here for `convert`?

softioc/device.py
```diff
@@ -1,9 +1,9 @@
 import os
+import sys
 import time
-import inspect
-from ctypes import *
 import numpy
 
+from ctypes import *
```
Araneidae, line 6: I tend to be a bit queasy about `import *`, and I suspect there aren't that many uses of `ctypes` values in this file, so try replacing this with a simple `import ctypes`.

AlexanderWells-diamond: There are 14 references to items in that file - I'll replace the import with 
`from ctypes import CFUNCTYPE, c_int, c_void_p, memmove`

Araneidae: Didn't realise that was inherited: see the line popped up because you moved it!

softioc/device.py
```diff
@@ -323,35 +375,43 @@ def _write_value(self, record, value):
...
+                value + b'\0',
+                dtype = numpy.uint8
+            )
 
-        value = numpy.require(value, dtype = self.dtype)
-        self._value = (+value, severity, alarm, timestamp)
-        if value.shape == ():
-            value.shape = (1,)
-        assert value.ndim == 1, 'Can\'t write multidimensional arrays'
+        return value
```
Araneidae, line 395: So what happens to values that don't pass either `isinstance` test?  Just passed through?

That's what the `numpy.require` was about, and it does the appropriate `.dtype` conversion here.

Or is that all subsumed into the `value_to_dbr` call that follows on?

Araneidae (commented):
Ok

softioc/device.py
```diff
@@ -323,35 +375,43 @@ def _write_value(self, record, value):
...
             # Convert a string into an array of characters.  This will produce
             # the correct behaviour when treating a character array as a string.
             # Note that the trailing null is needed to work around problems with
-            # some clients.
-            value = numpy.fromstring(value + '\0', dtype = 'uint8')
+            # some clients. Note this also exists in builder.py's _waveform().
+            value = numpy.frombuffer(
+                value + b'\0',
+                dtype = numpy.uint8
+            )
```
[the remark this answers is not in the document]
AlexanderWells-diamond: No good reason. This code probably used to be more indented thus requiring newlines. Fixing.

softioc/device.py
```diff
@@ -107,15 +171,28 @@ def __init__(self, name, **kargs):
 
         self.__validate = kargs.pop('validate', None)
         self.__always_update = kargs.pop('always_update', False)
-        self._value = kargs.pop('initial_value', None)
         self.__enable_write = True
+
+        if "initial_value" in kargs:
+            value = kargs.pop('initial_value')
+            if value is None:
+                raise ValueError("\"None\" is not a valid initial_value")
```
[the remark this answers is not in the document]
AlexanderWells-diamond: This depends if we care about a TypeError vs a ValueError exception. It's probably clear enough either way.
I suppose semantically we probably should push raising exceptions into the `_value_to_epics` method, so I will remove this check.

softioc/device.py
```diff
@@ -236,20 +266,37 @@ class GenericDevice(Base):
...
-bo = _Device_Out('bo', convert=False)
-stringin = _Device_In('stringin', mlst=False, default='')
-stringout = _Device_Out('stringout', mlst=False)
-mbbi = _Device_In('mbbi', convert=False)
-mbbo = _Device_Out('mbbo', convert=False)
+def epics_string(value):
+    """Trim a string to EPICS 40 (39 with null byte) character limit"""
+    if value is None:
+        raise ValueError("\"None\" not a valid value")
+
```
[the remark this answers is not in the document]
AlexanderWells-diamond: I took the indication that `None` was never a valid value quite literally and made all records reject it completely.

I think I prefer to leave this in, as it's the sort of error that may creep in during a running IOC and then they end up with some output string reporting "None", rather than an expected message. There's also then confusion if a user ever validly wants to write "None" as their value.
Then again, raising the exception is technically undocumented behaviour and so the IOC may well not catch it, and thus stop running altogether. What guidelines does PythonSoftIOC follow regarding trying to keep running in all cases?

softioc/device.py
AlexanderWells-diamond: Anything that isn't a `bytes` or a `str` will just pass through. This method is not supposed to touch them. I can move those checks out of the method if you prefer, although that does end up duplicating the `bytes` check. 

The diff makes this look pretty weird. The `numpy.require` does still happen, it's inside the `value_to_dbr` call. This method is just some pre-processing before that method.

softioc/device.py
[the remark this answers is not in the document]
AlexanderWells-diamond: After discussion I'll remove this special case check - there's so many other special cases that catching just this one isn't particularly worthwhile.

AlexanderWells-diamond:
The latest additions to the testware bring the number of tests up to ~500. This takes 5 minutes on my machine, and up to 15 minutes on a CI runner. This is getting to the stage where it's too long. There's also now non-trivial amounts of test code, to account for the wide variety of situations encountered (pythonSoftIOC issues, differing behaviour between In and Out records, differing behaviour of Waveforms, many individual tests skipped, etc.).

I'll look at refactoring these tests into larger tests that should hopefully run quicker (but not needing to spawn a new `multiprocess.process` for each test. This will probably take the form of having large numbers of records created per `process`. This'll likely lose resolution, as tests will likely stop on first failure rather than listing all failures.

AlexanderWells-diamond:
Tests have now been refactored to reduce the runtime. This was accomplished by making each subprocess run many records, rather than just one record per process. 

This has massively decreased runtime - down to 30 seconds from four and a half minutes on my machine, and a maximum of about 3 minutes across all CI platforms. However this does mean that test failures will be less specific, as tests will abort on the first incorrect record.

There's also been work on updating cibuildwheel to a more recent version. It seems that 2.3+ will not correctly build the `softioc/iocStatsDb` symlink into the built wheel on Windows. Attempts to restore the behaviour via pip environment variables didn't work, so I've pinned cibuildwheel 2.2.2.
See https://github.com/pypa/pip/issues/7555 for more info.

pyproject.toml
```diff
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools", "wheel", "setuptools_dso>=2.1", "epicscorelibs>=7.0.6.99.1.0"]
+requires = ["setuptools", "wheel", "setuptools_dso>=2.1", "epicscorelibs==7.0.6.99.1.0"]
```
Araneidae, line 2: This strong tie to a very particular version of `epicscorelibs` rings an alarm bell for me.  What's going on here?  Surely we don't want this, do we?

Separately, the version number is in an alarmingly non-standard format!

Araneidae (approved):
This is looking very good.

I did notice a number of new lines not covered by tests.  Given how comprehensive the tests already are now, I wonder if it would be worth adding a few more tests to cover these lines.

Finally, the documentation comment on byte strings makes me wonder whether we should have special support for `bytes`?  I think this is in just one place in the code: https://github.com/dls-controls/pythonSoftIOC/blob/98de8f675cb51bd85ffda88dd162dfdbd83de2a3/softioc/device.py#L322-L325 --- I'm ever so tempted to simply drop this special case!

pyproject.toml
thomascobb: > This strong tie to a very particular version of epicscorelibs rings an alarm bell for me. What's going on here?

There is an ordering issue here. The build process is this:
- deps from pyproject.toml are installed
- `python setup.py bdist_wheel` is called to make a binary wheel, which is compiled against the exact version of epicscorelibs that has been installed, and records this fact in the wheel
- the wheel is uploaded to PyPI
- anyone using this version of the binary wheel must have this exact version of epicscorelibs installed too (which pip will do automatically)
- anyone using the sdist is free to have whatever version of epicscorelibs they want

The `>=` that was there originally was correct, which would mean that pythonSoftIOC was compiled against the latest available epicscorelibs. Unfortunately in the tests we require p4p, which is also compiled against a particular version of epicscorelibs. If a new version of epicscorelibs is released, but p4p is not, then pip will fail to install dependencies. This is why we are using `==`. Without ABI compliance for EPICS, I can't see a way around this, unless a release of p4p is made with every epicscorelibs release. Fortunately we don't care for aioca as we only need API compliance, not ABI compliance.

@mdavidsaver can you think of a way we can resolve this?

> Separately, the version number is in an alarmingly non-standard format!

The first part is the EPICS release, but not sure about the other bits. Maybe @mdavidsaver could comment on this too?

mdavidsaver: > If a new version of epicscorelibs is released, but p4p is not

Sorry.  I've been sloppy about this as I'm in the (so far long) process of introducing another dependency [pvxslibs](https://pypi.org/project/pvxslibs/).  I'll try to straighten this out.  Though even if I do, there will always be some time when pypy.org has newer versions of parts of the dependency chain.  Maybe only minutes, but never zero.

> we only need API compliance, not ABI compliance.

Speaking strictly, API is important at build time (with a compiler), and ABI afterwards.  So in an ideal world, pyproject.toml is expressing the build time dependency (`BuildRequires` in an RPM .spec file), while `install_requires` in setup.py expresses the runtime dependency (plain `Requires` in a .spec).  As with a `.spec` runtime dependencies are partly generated.  so...

https://github.com/dls-controls/pythonSoftIOC/blob/3866562259982b6b019d3d5cb2b87b3c8ec98b01/setup.py#L108-L109

So if epicscorelibs `7.0.6.99.1.0` is installed, this will expand to:

```
epicscorelibs >=7.0.6.99.1.0, <7.0.6.99.2
```


> Separately, the version number is in an alarmingly non-standard format!

There are three pieces of information being encoded: upstream EPICS Base version (everything up to and including the `.99`), epicscorelibs ABI, and epicscorelibs revision.  So the version range being emitted above allows for later revisions which (hopefully) don't include any ABI change.

mdavidsaver: > @mdavidsaver can you think of a way we can resolve this?

If the problem is only with GHA, you might be able to pass `--no-binary p4p` to pip when installing the dev dependencies prior to running tests.

AlexanderWells-diamond: Thanks @mdavidsaver it looks like p4p 3.5.5 fixes the issue - I've restored the original epicscorelibs dependency in this commit and it seems to be working: https://github.com/dls-controls/pythonSoftIOC/pull/60/commits/adaa2868794c9f3281d159ba0e0d2cb427f4cd55

Araneidae:
@AlexanderWells-diamond , I'm happy for you to merge this when you're ready now.

PR #63, opened 2021-11-24
Fix main for Python2

AlexanderWells-diamond:
Fix the main function for Python2. When this is merged a new release should be done.

Note the version import change is not strictly required, it's just a fix for cases when `__main__` has been imported by someone else, which caused the import to pick up `softioc.py` instead of the module itself. This happens using `virtualenv` with its startup script. 

Closes #62 
Closes #59

PR #64, opened 2021-11-24
Remove python2 support

### LINKED ISSUE #61: Remove Python2 support
While PythonSoftIOC can support Python2, the code in `__main__.py` does not work in Python2.

As per discussion with @Araneidae and @thomascobb it's probably not worth fixing it, instead this issue will be to remove Python2 support entirely.

This will involve both a lot of code cleanup (there's a lot of Python version checks littered throughout the code) as well as reviewing any documentation about version support.

AlexanderWells-diamond:
This PR standardises Python 3.6 as the lowest supported version of Python, dropping support for Python 2.7.

3.6 was chosen as that's where our CI runs tests, although it wasn't declared as such in the setup.cfg until I edited it. If this should actually be Python3.7 we should also remove 3.6 from CI.

I have removed all references to 2.7, mostly by removing all Python version checking code, and done some fairly trivial refactoring where appropriate.

This also adds some auto-formatting of whitespace/newlines on save to VSCode, as we don't use an autoformatter. 

Closes #61


AlexanderWells-diamond:
This PR is parked until #63 is merged and released, then this branch will need to be rebased.

softioc/imports.py
```diff
@@ -34,27 +34,19 @@ def expect_true(status, function, args):
...
+    encoded = []
+    @classmethod
+    def from_param(cls, value):
+        if value is None:
+            return value
+        else:
+            return value.encode()
+
+def auto_decode(result, func, args):
+    return result.decode()
```
Araneidae, line 49: Looks like this function isn't actually being used, so can safely delete it.

Araneidae (commented):
Looks like we also have some Python version checks in `softioc/extension.c`; these can also go.

softioc/imports.py
AlexanderWells-diamond: Deleted.

AlexanderWells-diamond:
@Araneidae I've made all the changes requested, and CI is now passing.

It seems the previous error is a transient problem with codecov that disappeared on re-runs.

softioc/device_core.py
```diff
@@ -126,8 +126,7 @@ class DSET(DSET_BASE):
         # Hang onto the values we publish to EPICS to ensure that they persist!
         # We also need to ensure that the device name persits.
         cls.__dset = dset
-        if sys.version_info >= (3,):
-            cls._device_name_ = cls._device_name_.encode()
+        cls._device_name_ = cls._device_name_.encode()
         imports.registryDeviceSupportAdd(cls._device_name_, byref(cls.__dset))
```
Araneidae, line 130: Looking at how `_device_name_` is used, it's probably cleaner to rewrite these two lines as:
```python
       imports.registryDeviceSupportAdd(
            cls._device_name_.encode(), byref(cls.__dset))
```
The only reason this string was encoded was precisely for this method code.

Even better, how about changing the first field of `argtypes` for `registerDeviceSupportAdd` (in `imports.py`) from `c_char_p` to `auto_encode`?  Then we don't need to call `.encode()` at all!

softioc/imports.py
```diff
@@ -34,27 +34,16 @@ def expect_true(status, function, args):
...
-                return value
-            else:
-                return value.encode()
-
-    def auto_decode(result, func, args):
-        return result.decode()
+
+# Encode all strings to c_char_p
+class auto_encode(c_char_p):
+    encoded = []
```
Araneidae, line 40: I suspect the `encoded` variable was a mistake.  It certainly serves no purpose now and should be removed.

softioc/extension.c
```diff
@@ -226,27 +221,20 @@ static struct PyMethodDef softioc_methods[] = {
...
-#if PY_MAJOR_VERSION >= 3
-#  define PyMOD(NAME) PyObject* PyInit_##NAME (void)
-#else
-#  define PyMOD(NAME) void init##NAME (void)
-#endif
+
+#define PyMOD(NAME) PyObject* PyInit_##NAME (void)
+
 
 PyMOD(_extension)
```
Araneidae, line 237: Nah, don't use this ugly `PyMOD` macro, we don't need it anymore.  Just write
```python
PyObject *PyInit__extension(void)
```
here instead (and delete the `PyMOD` definition, of course).

Araneidae (commented):
Couple more tweaks, then we're there.

softioc/device_core.py
AlexanderWells-diamond: As discussed, we cannot do this as Python must keep hold of the data, to ensure the memory is not garbage-collected.

softioc/extension.c
AlexanderWells-diamond: Done.

softioc/imports.py
AlexanderWells-diamond: Done.

softioc/device_core.py
Araneidae: Yep :)

Araneidae:
Looks good to me.

Araneidae (approved):

PR #79, opened 2022-02-03
Allow NELM and length as aliases for waveform length

Araneidae:
Fixes issue #37

### DIFF
--- a/softioc/builder.py
+++ b/softioc/builder.py
@@ -134,6 +134,19 @@ DbfStringToNumpy = {
 }
 
 
+def _get_length(fields, default = None):
+    '''Helper function for getting 'length' or 'NELM' from arguments'''
+
+    if 'length' in fields:
+        assert 'NELM' not in fields, 'Cannot specify NELM and length together'
+        return fields.pop('length')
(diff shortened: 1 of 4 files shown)


AlexanderWells-diamond:
@Araneidae I think this is done now, I added two separate sets of tests in my two commits. 

I will note that I haven't managed to explain the failures I saw on CI, and the fixes I did for them were little more than adding a few more `assert` statements to check some return codes. Simply adding these made the failures disappear for no reason I can fathom. I'm slightly concerned that there's some underlying problem with the entire test framework I've written, perhaps something in `cothread` or `pythonSoftIOC` itself that really doesn't like being used alongside `multiprocessing`. Exactly what it may be I can't say unfortunately.

Happy for you to merge this whenever/however you think best.

AlexanderWells-diamond:
@Araneidae I think I've finished test fixes. My last commit has passed 10 consecutive times on the `test_fix` branch's CI. 

Let me know when/how you want this merged back to master.

Araneidae:
Looks like a clean commit chain that solves a lot of problems.

Go for it, and think we can, finally, go for a release!

Issue #81, opened 2022-02-14
Inconsistencies around defaults for Analog and Long records


AlexanderWells-diamond:
The fields which are defaulted for analogIn/Out and longIn/Out records in builder.py are inconsistent in various ways:

- aOut will default the DRVL and DEVH fields, but longOut will not
- long records specify EGU but analog records do not

The documentation also reflects these same inconsistencies and should be adjusted to whatever fix is made.

Araneidae:
I think it would be best if the fields matched the field definitions used in https://github.com/dls-controls/epics_device, specifically [the following definitions](https://github.com/dls-controls/epics_device/blob/1c717b32c6cce79cf1d621a4fb5199fb8f3f8f11/epics_device/device.py#L121-L139):
```python
def aIn(name, LOPR=None, HOPR=None, EGU=None, PREC=None, **fields):
def aOut(name, DRVL=None, DRVH=None, EGU=None, PREC=None, **fields):
def longIn(name, LOPR=None, HOPR=None, EGU=None, **fields):
def longOut(name, DRVL=None, DRVH=None, EGU=None, **fields):
```
with `LOPR` and `HOPR` automatically derived from `DRVL` and `DRVH` respectively if not already specified.

We can safely add the `EGU` and `PREC` fields where not already present without backwards compatibility issues, but there will be a potentially incompatible change affecting `aOut` if called in the following way:
```python
builder.aIn('NAME', -1, 1, DRVL=-1, DRVH=1)
```
(This will fail to run if the code is changed as suggested above.)

I'm inclined to accept this as a minor issue.

Issue #67, opened 2021-12-01
PINI flag is not set on In records when .set() is used


AlexanderWells-diamond:
In `builder.py`'s `_in_record` function we see this code:

```
if 'initial_value' in fields:
        fields.setdefault('PINI', 'YES')
```

There is no equivalent code for when `.set()` is called. This means that In records do not process on startup correctly such that `caget` sees the value. For example run the following IOC in a terminal:

```
# Import the basic framework components.
from softioc import softioc, builder, asyncio_dispatcher
import numpy
# Create an asyncio dispatcher, the event loop is now running
dispatcher = asyncio_dispatcher.AsyncioDispatcher()

# Set the record prefix
builder.SetDeviceName("PREFIX")

# Create some records
li = builder.longIn("TEST")

li.set(10)

# Boilerplate get the IOC started
builder.LoadDatabase()
softioc.iocInit(dispatcher)


# Finally leave the IOC running with an interactive shell.
softioc.interactive_ioc(globals())

```

And from another do `caget`:
```
[...]$ caget PREFIX:TEST
PREFIX:TEST                    0
```

And we see that the `.set()` has not taken effect.

PR #82, opened 2022-02-16
Rework the handling of default values for in and out records

Araneidae:
This commit addresses both github issues #81 and #67.

The arguments to the a and long record builder functions (`aIn`, `aOut`,
`longIn`, `longOut`) are brought more into line with the arguments as
used in the epics_device module.  This means that extra unnamed arguments
can be used to specify EGU and PREC (where appropriate), and for out
records the DRV limits take precedence.

As a side effect, the implementation of this is made closer to that in
https://github.com/dls-controls/epics_device/

### DIFF
--- a/softioc/builder.py
+++ b/softioc/builder.py
@@ -18,45 +18,52 @@ PythonDevice = pythonSoftIoc.PythonDevice()
 # The SCAN field for all input records defaults to I/O Intr.
 
 
-def _in_record(record, name, **fields):
-    '''For input records we provide some automatic extra features: scanning,
-    initialisation as appropriate, and blocking puts from outside the IOC.'''
-
+def _set_in_defaults(fields):
     fields.setdefault('SCAN', 'I/O Intr')
-    if 'initial_value' in fields:
-        fields.setdefault('PINI', 'YES')
+    fields.setdefault('PINI', 'YES')
     fields.setdefault('DISP', 1)
-    return getattr(PythonDevice, record)(name, **fields)
+
+def _set_out_defaults(fields):
+    fields.setdefault('OMSL', 'supervisory')
+
+# For longout and ao we want DRV{L,H} to match {L,H}OPR by default
+def _set_scalar_out_defaults(fields, DRVL, DRVH):
+    fields['DRVL'] = DRVL
+    fields['DRVH'] = DRVH
+    fields.setdefault('LOPR', DRVL)
+    fields.setdefault('HOPR', DRVH)
 
 
-def aIn(name, LOPR=None, HOPR=None, **fields):
-    return _in_record(
-        'ai', name, LOPR = LOPR, HOPR = HOPR, **fields)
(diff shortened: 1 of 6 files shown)


softioc/builder.py
```diff
@@ -18,45 +18,52 @@
...
-
+def _set_in_defaults(fields):
     fields.setdefault('SCAN', 'I/O Intr')
-    if 'initial_value' in fields:
-        fields.setdefault('PINI', 'YES')
+    fields.setdefault('PINI', 'YES')
     fields.setdefault('DISP', 1)
-    return getattr(PythonDevice, record)(name, **fields)
+
+def _set_out_defaults(fields):
```
AlexanderWells-diamond, line 26: What is the reason this isn't applied to `WaveformOut` and `longStringOut` records?

softioc/builder.py
```diff
@@ -18,45 +18,52 @@
...
+    fields.setdefault('OMSL', 'supervisory')
+
+# For longout and ao we want DRV{L,H} to match {L,H}OPR by default
+def _set_scalar_out_defaults(fields, DRVL, DRVH):
```
AlexanderWells-diamond, line 30: May just be my understanding, but "scalar" seems like the wrong word to use here? My assumption was this would apply to all non-Waveform Out records, as from my understanding they're all scalars.

AlexanderWells-diamond (commented):
Changes seem good to me. Just a couple of clarifying questions. 

There's no new tests, so I don't believe anything is currently exercising the `_set_scalar_out_defaults` function fully. It'd be good to see at least a trivial test that tries to set different DRVL/H and L/HOPR values and confirm the record is as expected. I don't know if there's any situations that might give invalid values (e.g. someone sets LOPR and HOPR the wrong way round) that may also warrant testing?

softioc/builder.py
AlexanderWells-diamond: Yes, looking through the manual agrees with that. That's part of the reason I find the name confusing - to me "scalar" means any single-valued record, but this function takes it to mean analog or long records exclusively. Is this an EPICS convention I'm not aware of?

Araneidae: I've looked up a couple of definitions of the word "scalar", and one of the defining properties of a "scalar" is that you can add values of this type (and multiply them, "scaling" them as it were).  It is not meaningful to add or scale booleans, enums or strings, so they're not "scalars"!

I just needed a simple shorthand name for this action; could have been `_set_drvl_drvh()` or similar.

softioc/builder.py
Araneidae: The `waveform` record doesn't have an `OMSL` field, and as it happens that's the only field that `_set_out_defaults()` sets.

It's an anomaly, you'd expect an out `waveform` to be treated like any other Out record ... but in truth, it's _not_ an Out record at all, it's a `waveform` record (which can work in either direction) with special treatment.

Araneidae:
Regarding the lack of tests, I agree that it would be good to add some more tests.

AlexanderWells-diamond (approved):

PR #98, opened 2022-07-28
Caput with callback

AlexanderWells-diamond:
Add blocking to the processing of `on_update`/`on_update_name` callbacks, to allow `caput` to correctly wait for processing to complete before returning.

Documented the new `blocking` flag on OUT record creation, and the `SetBlocking` function. I didn't document the CothreadDispatcher, as there's no reason for any users to explicitly create it themselves. Let me know if I should document it, or perhaps change its name to indicate its internal-only.

Tests written:
- Various tests that the record-level and global blocking flag are correctly passed to the `ProcessDeviceSupportOut` class
- A blocking record will correctly process blocking `caput` requests from both a single thread multiple times and from multiple threads simultaneously.

### DIFF
--- a/softioc/asyncio_dispatcher.py
+++ b/softioc/asyncio_dispatcher.py
@@ -32,18 +32,19 @@ class AsyncioDispatcher:
         else:
             self.loop = loop
 
-    def __call__(self, func, completion, func_args=(), completion_args=()):
+    def __call__(
+            self,
+            func,
+            func_args=(),
+            completion = None,
+            completion_args=()):
         async def async_wrapper():
             try:
-                if inspect.iscoroutinefunction(func):
-                    await func(*func_args)
-                else:
-                    ret = func(*func_args)
-                    # Handle the case of a synchronous function that returns a
-                    # coroutine, like the lambda for on_update_name does
-                    if inspect.isawaitable(ret):
-                        await ret
-                completion(*completion_args)
+                ret = func(*func_args)
+                if inspect.isawaitable(ret):
+                    await ret
+                if completion:
+                    completion(*completion_args)
             except Exception:
                 logging.exception("Exception when awaiting callback")
         asyncio.run_coroutine_threadsafe(async_wrapper(), self.loop)
--- a/softioc/asyncio_dispatcher.py
+++ b/softioc/asyncio_dispatcher.py
@@ -32,12 +32,18 @@ class AsyncioDispatcher:
         else:
             self.loop = loop
 
-    def __call__(self, func, *args):
+    def __call__(self, func, completion, func_args=(), completion_args=()):
         async def async_wrapper():
             try:
-                ret = func(*args)
-                if inspect.isawaitable(ret):
-                    await ret
+                if inspect.iscoroutinefunction(func):
+                    await func(*func_args)
+                else:
+                    ret = func(*func_args)
+                    # Handle the case of a synchronous function that returns a
+                    # coroutine, like the lambda for on_update_name does
+                    if inspect.isawaitable(ret):
+                        await ret
+                completion(*completion_args)
             except Exception:
                 logging.exception("Exception when awaiting callback")
         asyncio.run_coroutine_threadsafe(async_wrapper(), self.loop)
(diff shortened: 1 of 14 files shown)


softioc/device.py
```diff
@@ -1,19 +1,36 @@
...
 dispatcher = None
 
+# Global blocking flag, used to mark asynchronous (False) or synchronous (True)
+# processing modes for Out records.
+# Default False to maintain behaviour from previous versions.
+blocking = False
+
+def SetBlocking(val):
+    global blocking
+    blocking = val
```
Araneidae, line 32: It wouldn't hurt to return the old value of `blocking`.  This would then allow a coding style of the form:
```python
old_blocking = SetBlocking(new_blocking)
create_records()
SetBlocking(old_blocking)
```
which could even be wrapped up as a `with` using a context manager.

Not suggesting we write the context manager, just that we return the old state!

softioc/device.py
```diff
@@ -162,9 +184,19 @@ def init_record(self, record):
             recGblResetAlarms(record)
         return self._epics_rc_
 
+    def __completion(self, record):
+        '''Signals that all on_update processing is finished'''
+        if self._blocking:
+            signal_processing_complete(record, self._callback)
+        pass
```
Araneidae, line 191: Forgotten relic!

softioc/device.py
```diff
@@ -183,7 +215,14 @@ def _process(self, record):
...
             record.UDF = 0
             if self.__on_update and self.__enable_write:
-                dispatcher(self.__on_update, python_value)
+                record.PACT = self._blocking
+                dispatcher(
+                    self.__on_update,
+                    self.__completion,
+                    func_args=(python_value,),
+                    completion_args=(record,)
+                )
```
Araneidae, line 224: Not a huge fan of the dangling `)` here, FWIW.

Araneidae: After all the pain, this has come out remarkably cleanly!

softioc/imports.py
```diff
@@ -27,6 +27,17 @@ def install_pv_logging(acf_file):
...
+def create_callback_capsule():
+    return _extension.create_callback_capsule()
+
+def signal_processing_complete(record, callback):
+    '''Signal that asynchronous record processing has completed'''
+    _extension.signal_processing_complete(
+        record.PRIO,
+        record.record.value,
+        callback
+    )
```
Araneidae, line 39: Am sure this dangling `)` style is new to this code!

softioc/asyncio_dispatcher.py
```diff
@@ -32,12 +32,18 @@ def aioJoin(worker=worker, loop=self.loop):
...
-                if inspect.isawaitable(ret):
-                    await ret
+                if inspect.iscoroutinefunction(func):
+                    await func(*func_args)
+                else:
+                    ret = func(*func_args)
+                    # Handle the case of a synchronous function that returns a
+                    # coroutine, like the lambda for on_update_name does
+                    if inspect.isawaitable(ret):
+                        await ret
```
Araneidae, line 45: Hmm.  Suspect this would be more simply written as
```python
ret = func(*func_args)
if inspect.isawaitable(ret):
    await ret
```

Araneidae (approved):
Looks pretty good.  Have a couple of coding changes to suggest, particularly the asyncio dispatcher, but I think this is a huge success.

After all the pain this has come out a lot more cleanly than I was expecting!

softioc/device.py
AlexanderWells-diamond: Added in a0bce0ce330bb96274d4bbebff179c872165d7e6

softioc/device.py
AlexanderWells-diamond: Fixed in a0bce0ce330bb96274d4bbebff179c872165d7e6

softioc/imports.py
AlexanderWells-diamond: Fixed in a0bce0ce330bb96274d4bbebff179c872165d7e6

softioc/device.py
AlexanderWells-diamond: Fixed in a0bce0ce330bb96274d4bbebff179c872165d7e6

softioc/asyncio_dispatcher.py
AlexanderWells-diamond: Fixed in a0bce0ce330bb96274d4bbebff179c872165d7e6

PR #102, opened 2022-08-25
Allow arrays of strings to be used with Waveforms

AlexanderWells-diamond:
Allow creation records like:
```
t1 = builder.WaveformOut(
    "TEST1",  initial_value=["ZERO", "ONE", "MANY"]
)
t2 = builder.WaveformOut(
    "TEST2",  initial_value=[b"ZERO", b"ONE", b"MANY"]
)
```

Where each element of the array is treated as an EPICS string

Values returned will always be Python string lists:

```
> t1.get()
["ZERO", "ONE", "MANY"]
> t2.get()
["ZERO", "ONE", "MANY"]
```

(valid) Unicode characters may also be passed in, and will be returned as-is. Note that it is not always obvious how many Unicode characters can be safely entered into EPICS's 40 character string limits.

### DIFF
--- a/docs/reference/api.rst
+++ b/docs/reference/api.rst
@@ -342,6 +342,11 @@ All functions return a wrapped `ProcessDeviceSupportIn` or
     field type name.  Otherwise the field type is taken from the initial value
     if given, or defaults to ``'FLOAT'``.
 
+    .. note::
+        When storing arrays of strings, it is possible to store Unicode characters.
+        However, as EPICS has no Unicode support the resultant values will be stored
+        as byte strings. Care must be taken when encoding/decoding the values.
+
 
 The following functions generates specialised records.
 
--- a/docs/reference/api.rst
+++ b/docs/reference/api.rst
@@ -343,9 +343,9 @@ All functions return a wrapped `ProcessDeviceSupportIn` or
     if given, or defaults to ``'FLOAT'``.
 
     .. note::
-        When storing arrays of strings, it is possible to store Unicode characters.
-        However, as EPICS has no Unicode support the resultant values will be stored
-        as byte strings. Care must be taken when encoding/decoding the values.
+        Storing arrays of strings differs from other values. String arrays will always
+        be assumed to be encoded as Unicode strings, and will be returned to the user
+        as a Python list rather than a Numpy array.
 
 
 The following functions generates specialised records.
--- a/softioc/builder.py
+++ b/softioc/builder.py
@@ -11,7 +11,7 @@ LoadDbdFile(os.path.join(os.path.dirname(__file__), 'device.dbd'))
 
 from . import device, pythonSoftIoc  # noqa
 # Re-export this so users only have to import the builder
-from .device import SetBlocking # noqa
+from .device import SetBlocking, to_epics_str_array # noqa
 
 PythonDevice = pythonSoftIoc.PythonDevice()
 
@@ -148,7 +148,7 @@ NumpyDtypeToDbf = {
     'uint32':   'ULONG',
     'float32':  'FLOAT',
     'float64':  'DOUBLE',
-    'bytes320': 'STRING',  # Numpy's term for a 40-character string (40*8 bits)
+    'bytes320': 'STRING',  # Numpy term for 40-character byte str (40*8 bits)
 }
 
 # Coverts FTVL string to numpy type
@@ -220,8 +220,8 @@ def _waveform(value, fields):
                 initial_value = numpy.require(initial_value, numpy.int32)
             elif initial_value.dtype == numpy.uint64:
                 initial_value = numpy.require(initial_value, numpy.uint32)
-            elif initial_value.dtype.char in ("S", "U"):
-                initial_value = numpy.require(initial_value, numpy.dtype("S40"))
+            elif initial_value.dtype.char in ('S', 'U'):
+                initial_value = to_epics_str_array(initial_value)
     else:
         initial_value = numpy.array([], dtype = datatype)
         length = _get_length(fields)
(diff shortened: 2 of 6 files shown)


AlexanderWells-diamond:
The current state of the branch is that you can create a WaveformIn/Out with an initial_value of an array of (byte) strings, but you cannot set that value at a later time. This is because if there is no `dtype`, `numpy` defaults to `float64`. 

If you specify an `FTVL` or `dtype` of the relevant type, this does behave as expected.

I'm unsure if we can support arrays of strings quite as freely as we could in version 3.2.1 - I don't see a neat way around the implicit `numpy` conversion.

softioc/builder.py
```diff
@@ -195,11 +198,15 @@ def _waveform(value, fields):
...
         # Special case for [u]int64: if the initial value comes in as 64 bit
         # integers we cannot represent that, so recast it as [u]int32
+        # Special case for array of strings to correctly identify each element
+        # of the array as a string type.
         if datatype is None:
             if initial_value.dtype == numpy.int64:
                 initial_value = numpy.require(initial_value, numpy.int32)
             elif initial_value.dtype == numpy.uint64:
                 initial_value = numpy.require(initial_value, numpy.uint32)
+            elif initial_value.dtype.char == "S":
```
coretl, line 208:
```suggestion
            elif initial_value.dtype.char in ("S", "U"):
```
Might need to catch if people pass List[str] rather than List[bytes] in...

AlexanderWells-diamond: I've added this case, as well as a simple test without Unicode characters. I'm not sure what the correct result for Unicode characters is; should it end up with the bytes representation of the character as a "byte string" or as an array of numbers?

tests/test_record_values.py
```diff
@@ -31,6 +31,9 @@
     "might think to put into a record that can theoretically hold a huge " \
     "string and so lets test it and prove that shall we?"
 
+# The numpy dtype of all arrays of strings
+NUMPY_DTYPE_STRING = "S40"
```
Araneidae, line 35: Don't think you use this anymore

docs/reference/api.rst
```diff
@@ -342,6 +342,11 @@ All functions return a wrapped `ProcessDeviceSupportIn` or
     field type name.  Otherwise the field type is taken from the initial value
     if given, or defaults to ``'FLOAT'``.
 
+    .. note::
+        When storing arrays of strings, it is possible to store Unicode characters.
+        However, as EPICS has no Unicode support the resultant values will be stored
+        as byte strings. Care must be taken when encoding/decoding the values.
+
```
Araneidae, line 349: Don't really think this note captures the essence of this new feature.  In particular, the internal encoding is almost completely hidden from users, and it's a bit of a misconception to say that "EPICS has no Unicode support".

We should reference the fact that waveforms of strings are treated as Python Unicode strings throughout and are managed as Python arrays unlike all other waveforms.

softioc/device.py
```diff
@@ -351,11 +351,29 @@ class ao(ProcessDeviceSupportOut):
...
+    nul-terminated strings"""
+    result = numpy.empty(len(value), 'S40')
+
+    for n, s in enumerate(value):
+        if isinstance(s, str):
+            val = EpicsString._ctype_()
+            val.value = s.encode() + b'\0'
+            result[n] = val.value
+        else:
+            result[n] = s
```
Araneidae, line 365: Hmm.  Does this do the right thing when `s` is a strange value?

softioc/device.py
```diff
@@ -351,11 +351,29 @@ class ao(ProcessDeviceSupportOut):
...
+
 
 def _require_waveform(value, dtype):
     if isinstance(value, bytes):
         # Special case hack for byte arrays.  Surprisingly tricky:
         value = numpy.frombuffer(value, dtype = numpy.uint8)
+
+    if dtype and dtype.char == 'S':
+        return to_epics_str_array(value)
+
```
Araneidae, line 376: Not a huge fan of early return, and looking at the context this is a case for using `else:`.

Also, this test should probably be done _before_ the `bytes` check above, making this a stronger case for being an `if ... else ...` block.

Araneidae (changes requested):
Think there are a few minor things to tidy up, but looks good.

softioc/builder.py
```diff
@@ -132,6 +132,8 @@ def Action(name, **fields):
     'uint32':   'ULONG',
     'float32':  'FLOAT',
     'float64':  'DOUBLE',
+    'bytes32':  'STRING',
+    'bytes320': 'STRING',
```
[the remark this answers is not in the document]
Araneidae: Ok, we've discussed this.  Not sure they're _still_ required after other changes, worth checking, and you have commented.  Resolving as outdated.

tests/test_record_values.py
AlexanderWells-diamond: Removed in 5b59ed7f86a126842467247dfb004ecf21aed561

docs/reference/api.rst
AlexanderWells-diamond: Rewritten in 5b59ed7f86a126842467247dfb004ecf21aed561 - how does that sound?

softioc/device.py
AlexanderWells-diamond: I added some tests for mixed lists in 5b59ed7f86a126842467247dfb004ecf21aed561 - i.e. mixing strings and integers in a single list.

softioc/device.py
AlexanderWells-diamond: Rewritten in 5b59ed7f86a126842467247dfb004ecf21aed561

Araneidae (approved):
ok

PR #111, opened 2022-10-13
Allow "status" and "severity" on In record init

### LINKED ISSUE #57: Allow setting of alarm status and severity on record creation using PythonSoftIOC alarm values
Currently to set the alarm status and severity on a record during creation you must use the EPICS field names and values, e.g.:

`STAT="UDF", SEVR="INVALID"`

It would be convenient to allow use of the same syntax that the `.set_alarm()` method uses, that is to say making use of the `alarm` enum. 

This is probably somewhat related to #53.

AlexanderWells-diamond:
This PR allows users to specify `severity` and `status` on the command line, with values from the `softioc.alarm` module. 

Closes #57

### DIFF
--- a/softioc/builder.py
+++ b/softioc/builder.py
@@ -1,5 +1,6 @@
 import os
 import numpy
+
 from .softioc import dbLoadDatabase
 
 from epicsdbbuilder import *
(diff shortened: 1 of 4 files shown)


softioc/builder.py
```diff
@@ -24,10 +25,17 @@ def _set_in_defaults(fields):
...
 
 def _set_out_defaults(fields):
     fields.setdefault('OMSL', 'supervisory')
 
+def _set_alarm(fields):
+    if "status" in fields:
+        fields['STAT'] = _statStrings[fields.pop('status')]
+    if "severity" in fields:
+        fields['SEVR'] = _severityStrings[fields.pop('severity')]
+
```
Araneidae, line 38: Might be good to fail at this point of `'STAT'` or `'SEVR'` is already set.  Can't think of an easy way, but the following helper function would do the job:
```python
def _set_field(fields, name, value):
    assert name not in fields, 'Field %s already defined' % name
    fields[name] = value
```

Araneidae (approved):
Looks pretty straightforward.  Probably worth making the one mentioned change.

softioc/builder.py
AlexanderWells-diamond: I've committed a slight alternative solution, based on similar logic where we already disallow certain keyword pairs. I think this is more in keeping with similar situations, and will provide a clearer error message.

PR #114, opened 2022-10-18
Allow users to reset the list of created records

### LINKED ISSUE #56: Allow users to reset the list of created records
`epicsdbbuilder` exports the method `ResetRecords()`, [documentation here](https://dls-controls.github.io/epicsdbbuilder/master/api.html?highlight=reset#epicsdbbuilder.ResetRecords), which allows the known list of records to be forgotten.

PythonSoftIOC maintains its own dictionary of records in `RecordLookup`, which appears to offer the same protections are epicsdbbuilder but offers no reset mechanism. 

The situation I'm in is that I'm trying to test my record creation (note: not actually starting the IOC, just testing the creation). An `epicsdbbuilder` error about duplicat records is encountered first, but calling `ResetRecords()` allows me to work around that, within the published API. However I have no (sensible) way to reset the dictionary that PythonSoftIOC maintains. This would force me to create a multi-process test framework, which in quick tests is 2x-3x slower than running in a single process.

AlexanderWells-diamond:
Implemented a "ClearRecords" function that performs logic that was previously only in some code for tests.

Closes #56

### DIFF
--- a/docs/reference/api.rst
+++ b/docs/reference/api.rst
@@ -428,7 +428,9 @@ The following attributes allow more direct access to record creation.
 
 .. function:: ClearRecords()
     This can be used to remove all created records. This means the same record
-    names can be re-used for different record types.
+    names can be re-used for different record types. This cannot be used once the
+    record database has been loaded, and is only designed to be used to help with
+    testing.
 
 Finally, the following function is used to load record definitions before
 starting the IOC.
--- a/docs/reference/api.rst
+++ b/docs/reference/api.rst
@@ -426,6 +426,10 @@ The following attributes allow more direct access to record creation.
 
     See `../how-to/use-soft-records` for a full example of its usage.
 
+.. function:: ClearRecords()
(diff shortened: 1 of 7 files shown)


docs/reference/api.rst
```diff
@@ -426,6 +426,10 @@ The following attributes allow more direct access to record creation.
 
     See `../how-to/use-soft-records` for a full example of its usage.
 
+.. function:: ClearRecords()
+    This can be used to remove all created records. This means the same record
+    names can be re-used for different record types.
+
```
Araneidae, line 432: I would expand on this a little, just to say:

> This cannot be used once the record database has been loaded, and is only designed to be used to help with testing.

Araneidae (approved):
Looks good to me

docs/reference/api.rst
AlexanderWells-diamond: Done in c422b3404a2b25435c41a4f4c82da8dd9731ca4b

PR #138, opened 2023-07-16
AsyncioDispatcher cleanup tasks atexit

mdavidsaver:
Currently, the `atexit` logic simply breaks the event loop leaving an in-progress `Task` incomplete.  Instead, use `asyncio.run()`, which cancels tasks before returning (among other cleanup).

Motivated by usage of `asyncio.create_subprocess_exec()` where the present behavior leaves the child process running.  Similarly, for cleanup of temporary files/directories.

The downside is that `asyncio.run()` is introduced in python 3.7.  So the 3.6 tests will fail.  Emulating `asyncio.run()`  in 3.6 would be possible, but is it worth the effort?

I don't know what is going wrong with the windows and mac tests.

### DIFF
--- a/softioc/asyncio_dispatcher.py
+++ b/softioc/asyncio_dispatcher.py
@@ -5,32 +5,60 @@ import threading
 import atexit
 
 class AsyncioDispatcher:
-    def __init__(self, loop=None):
+    def __init__(self, loop=None, debug=False):
         """A dispatcher for `asyncio` based IOCs, suitable to be passed to
         `softioc.iocInit`. Means that `on_update` callback functions can be
         async.
 
         If a ``loop`` is provided it must already be running. Otherwise a new
         Event Loop will be created and run in a dedicated thread.
+        ``debug`` is passed through to ``asyncio.run()``.
+
+        For a clean exit, call ``softioc.interactive_ioc(..., call_exit=False)``
         """
         if loop is None:
+            # will wait until worker is executing the new loop
+            started = threading.Event()
             # Make one and run it in a background thread
-            self.loop = asyncio.new_event_loop()
-            worker = threading.Thread(target=self.loop.run_forever)
+            self.__worker = threading.Thread(target=asyncio.run,
+                                             args=(self.__inloop(started),),
+                                             kwargs={'debug': debug})
             # Explicitly manage worker thread as part of interpreter shutdown.
             # Otherwise threading module will deadlock trying to join()
             # before our atexit hook runs, while the loop is still running.
-            worker.daemon = True
+            self.__worker.daemon = True
 
-            @atexit.register
-            def aioJoin(worker=worker, loop=self.loop):
-                loop.call_soon_threadsafe(loop.stop)
-                worker.join()
-            worker.start()
+            self.__worker.start()
+            started.wait()
+
+            self.__atexit = atexit.register(self.__shutdown)
+
+            assert self.loop is not None and self.loop.is_running()
+
         elif not loop.is_running():
             raise ValueError("Provided asyncio event loop is not running")
         else:
             self.loop = loop
+
+    def close(self):
+        if self.__atexit is not None:
+            atexit.unregister(self.__atexit)
+            self.__atexit = None
+
+        self.__shutdown()
+
+    async def __inloop(self, started):
+        self.loop = asyncio.get_running_loop()
+        self.__interrupt = asyncio.Event()
+        started.set()
+        del started
+        await self.__interrupt.wait()
+
+    def __shutdown(self):
+        if self.__worker is not None:
+            self.loop.call_soon_threadsafe(self.__interrupt.set)
+            self.__worker.join()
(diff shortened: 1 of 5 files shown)


softioc/asyncio_dispatcher.py
```diff
@@ -5,33 +5,61 @@
...
+            atexit.unregister(self.__atexit)
+            self.__atexit = None
+
+        self.__shutdown()
+
+    async def __inloop(self, started):
+        self.loop = asyncio.get_running_loop()
+        self.__interrupt = asyncio.Event()
+        started.set()
+        del started
```
coretl, line 54: Why do you have to del the Event here?

softioc/asyncio_dispatcher.py
```diff
@@ -48,3 +76,9 @@ async def async_wrapper():
             except Exception:
                 logging.exception("Exception when running dispatched callback")
         asyncio.run_coroutine_threadsafe(async_wrapper(), self.loop)
+
+    def __enter__(self):
+        return self
```
coretl, line 81: Did you have a usecase for having the dispatcher as a context manager?

coretl (commented):
I think it is worth ditching 3.6 so we get `asyncio.run`. @AlexanderWells-diamond @Araneidae what do you think? This looks ok to me, I asked a couple of questions out of interest. It need 3.6 ditching from the matrix before merging. The mac and windows tests seem to be working in the latest run?

AlexanderWells-diamond:
I have no attachment to Python 3.6 so am happy to see it go.

Araneidae:
@coretl, I also have no ties to 3.6 ... _except_ I would point out that it is the stock version of Python installed on RHEL8, so dropping it may have consequences for users on that platform.

softioc/asyncio_dispatcher.py
mdavidsaver: This is not required.  Just some housekeeping since this Event will not be reused.

mdavidsaver:
> ... stock version of Python installed on RHEL8 ...

This is a fair point.  I can think of two possibilities in addition to drop support for 3.6 or a full backport of `asyncio.run`.

State that only the asyncio part is only supported with py >= 3.7.  (fyi. driven by removal of `loop=` P4P only supports asyncio with py >= 3.6, but supports thread/cothread with py 2.7 and >= 3.4)

Continue the current behavior with <= 3.6

coretl:
My preference is to drop 3.6, bump the major release number, and see if anyone complains... Any objections?

AlexanderWells-diamond:
No objections here as long as its clear in the Changelog.

mdavidsaver:
Is there anything I can do to move this PR forward?

coretl:
> Is there anything I can do to move this PR forward?

If you can drop 3.6 from the GitHub actions tests matrix, then add a test with your usecase (checking a spawned subprocess running as a Task exits cleanly outside context manager) then I'd be happy to merge this.

@AlexanderWells-diamond @Araneidae any thoughts?

Araneidae:
I have learnt that there is a python3.11 rpm available on our RHEL8 channel and so presumably potentially available on all administered RHEL8 systems, so I agree that 3.6 can now go.

AlexanderWells-diamond:
This PR will need some tidying up before we can merge it (remove 3.6, documentation, changelog etc.); I'll do that in the near future.

AlexanderWells-diamond:
I've updated this PR to remove Python 3.6 support and added a Changelog entry.

softioc/asyncio_dispatcher.py
```diff
@@ -5,33 +5,61 @@
...
-            worker.start()
         elif not loop.is_running():
             raise ValueError("Provided asyncio event loop is not running")
         else:
             self.loop = loop
 
+    def close(self):
+        if self.__atexit is not None:
+            atexit.unregister(self.__atexit)
+            self.__atexit = None
```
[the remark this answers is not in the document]
coretl: My understanding:
- In IOC code we would normally just let the atexit handler catch it
- In tests then we would normally use the context manager
- There may be some tests where a context manager is inconvenient, then we would use the close method

AlexanderWells-diamond:
With my addition of a test for the Context Manager, I'm finished with my changes. @Araneidae and/or @coretl please merge when you are happy with it.

coretl (approved):

EmilioPeJu (approved):

softioc/asyncio_dispatcher.py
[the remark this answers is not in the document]
mdavidsaver: > * There may be some tests where a context manager is inconvenient, then we would use the close method

Or perhaps wanting to wrap this dispatcher object in another.  Or maybe just an artifact of my tendency to think of `with ... as X: ...` as shorthand for a `try: ... finally: X.close()`.

mdavidsaver:
@AlexanderWells-diamond Thank you for taking the time to finish this up!

Issue #153, opened 2024-03-11
Bug when Out records self-update on a Raspberry Pi


codedump:
Hello,

there appears to be a bug when using pythonSoftIOC on a Raspberry Pi.

Try out the "hello world" IOC example from the documentation (reproduced below without all the comments and without the `aIn` part, for better overview); then modify the `on_update=...` parameter to self-update the `aOut` record.

This is something that is necessary, for instance, to implement specific EPICS motor fields, e.g. `HOMF`, which are expected to update their own value back to 0 once they're done.

Here is the IOC in question:
```
from softioc import softioc, builder
import cothread
builder.SetDeviceName("MY-DEVICE-PREFIX")
ao = builder.aOut('AO', initial_value=12.45, always_update=True, on_update=lambda v: ao.set(7.62))
builder.LoadDatabase()
softioc.iocInit()
softioc.interactive_ioc(globals())
```

Then trying to actually trigger the self-update, by simply just writing a value -- any value:
```
$ caput MY-DEVICE-PREFIX:AO 3.14
```

This is how it fails on a Raspberry Pi v4, with fairly recent Debian 12.1 (I think it's Bookworm):

```
INFO: PVXS QSRV2 is loaded, permitted, and ENABLED.
Starting iocInit
############################################################################
## EPICS 7.0.7.0
## Rev. 7.0.7.99.0.2
## Rev. Date 7.0.7.99.0.2
############################################################################
iocRun: All initialization complete
Python 3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> Asynchronous callback raised uncaught exception
Traceback (most recent call last):
  File "/home/specuser/fifer-venv/lib/python3.11/site-packages/cothread/cothread.py", line 964, in callback_events
    action(*args)
  File "/home/specuser/fifer-venv/lib/python3.11/site-packages/softioc/cothread_dispatcher.py", line 29, in wrapper
    func(*func_args)
  File "/home/specuser/tmp/./foo.py", line 6, in <lambda>
    on_update=lambda v: ao.set(7.62))
                        ^^^^^^^^^^^^
  File "/home/specuser/fifer-venv/lib/python3.11/site-packages/softioc/device.py", line 264, in set
    db_put_field(_record.NAME, dbf_code, data, length)
[...]

PR #154, opened 2024-03-12
Fix conversion of ctypes pointers passed to C extension

Araneidae:
As reported in issue #153 there appears to be a problem in converting a Python integer generated by `ctypes.address()` back to a void* pointer when naively using `PyArg_ParseTuple`.


AlexanderWells-diamond (approved):
It needs an entry in `CHANGELOG.rst`, but otherwise it looks fine to me.

PR #157, opened 2024-04-08
Enable setting alarm status of Out records

### LINKED ISSUE #53: Be able to force Out records into an error state
Currently there is no explicit way to force an Out record into an error state. The `set()` method on In records provides the `severity` and `alarm` keyword arguments (as well as the `set_alarm()` method), but there is no equivalent mechanism for Out records. Providing this would be useful for the PandABlocks-client PythonSoftIOC integration, as PandA can report fields as being in_error, and the associated records are often Out types.

AlexanderWells-diamond:
Adds the `set_alarm` method to `ProcessDeviceSupportOut`, along with extra keyword arguments to `set`. As these are keyword args with defaults, there should be no compatibility problems.

Tests added to check calling `set_alarm()` before or after IocInit() causes the severity and status attributes to be set as expected.

Closes #53.

### DIFF
--- a/softioc/device.py
+++ b/softioc/device.py
@@ -137,7 +137,7 @@ class ProcessDeviceSupportIn(ProcessDeviceSupportCore):
         return self._epics_rc_
 
     def set(self, value,
-            severity=alarm.NO_ALARM, alarm=alarm.UDF_ALARM, timestamp=None):
+            severity=alarm.NO_ALARM, alarm=alarm.NO_ALARM, timestamp=None):
         '''Updates the stored value and triggers an update.  The alarm
         severity and timestamp can also be specified if appropriate.'''
         value = self._value_to_epics(value)
@@ -177,14 +177,16 @@ class ProcessDeviceSupportOut(ProcessDeviceSupportCore):
 
         if 'initial_value' in kargs:
             value = self._value_to_epics(kargs.pop('initial_value'))
-            initial_alarm = alarm.NO_ALARM
+            initial_severity = alarm.NO_ALARM
+            initial_status = alarm.NO_ALARM
         else:
             value = None
             # To maintain backwards compatibility, if there is no initial value
             # we mark the record as invalid
-            initial_alarm = alarm.INVALID_ALARM
+            initial_severity = alarm.INVALID_ALARM
+            initial_status = alarm.UDF_ALARM
 
-        self._value = (value, initial_alarm, alarm.UDF_ALARM)
+        self._value = (value, initial_severity, initial_status)
 
         self._blocking = kargs.pop('blocking', blocking)
         if self._blocking:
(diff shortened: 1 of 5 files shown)


AlexanderWells-diamond:
I've done a few more tweaks, and changed the default Status value to `NO_ALARM`, which is technically a breaking change we may need to highlight (although maybe not as a non-zero STAT is fairly meaningless with a SEVR of 0)

I have found one problem: Trying to set STAT and SEVR using the EPICS attribute names during record declaration doesn't work i.e.:
```
ai = builder.aIn('AI', SEVR="MINOR", STAT="HIHI")
ao = builder.aOut('AO', SEVR="MINOR", STAT="HIHI")
```
The records created here will not respect these SEVR and STAT values:
```
$ caget -a AI
MY-DEVICE-PREFIX:AI            2024-04-08 14:04:54.829984 0  
$ caget -a MY-DEVICE-PREFIX:AO
MY-DEVICE-PREFIX:AO            2024-04-08 14:04:54.328884 0 UDF INVALID
```

I believe this has never worked; they are passed to the `epicsdbbuilder` (as proven by giving invalid values producing an error message), but for reasons unknown at this time they are gone after record initialization.

Is it worth trying to support these keywords properly?

softioc/device.py
```diff
@@ -265,11 +288,11 @@ def set(self, value, process=True):
             self.__enable_write = True
 
     def get(self):
-        if self._value is None:
+        if self._value[0] is None:
             # Before startup complete if no value set return default value
             value = self._default_value()
         else:
-            value = self._value
+            value = self._value[0]
```
Araneidae, line 295: Can simplify this, `_value[0]` is now never `None`, can now just be
```python
return self._epics_to_value(self._value[0])
```

softioc/device.py
```diff
@@ -216,9 +223,14 @@ def _process(self, record):
         if record.PACT:
             return EPICS_OK
 
+        # Ignore memoized value, retrieve it from the VAL field directly later
+        _, severity, alarm = self._value
+
+        self.process_severity(record, severity, alarm)
+
         value = self._read_value(record)
```
Araneidae, line 231: Put reading `self._value` and `self._read_value(record)` on adjacent lines, then it's much clearer that they're working together and you might not need the comment!

Araneidae (approved):
Looks pretty good to me

softioc/device.py
```diff
@@ -176,9 +176,15 @@ def __init__(self, name, **kargs):
...
+            value = self._value_to_epics(kargs.pop('initial_value'))
+            initial_alarm = alarm.NO_ALARM
         else:
-            self._value = None
+            value = None
+            # To maintain backwards compatibility, if there is no initial value
+            # we mark the record as invalid
+            initial_alarm = alarm.INVALID_ALARM
+
+        self._value = (value, initial_alarm, alarm.UDF_ALARM)
```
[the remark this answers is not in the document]
AlexanderWells-diamond: Fixed in 5f461327c2190f77a72935bce0bb40e4d59337a3

PR #158, opened 2024-04-24
Fix CI for MacOS and Numpy2.0

AlexanderWells-diamond:
The current macos-latest tag points to macos-14, running on arm64. We currently can't support this architecture as it does not support Python3.7.

The current supported tags/runners is documented here: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

This is not a permanent fix and should be re-visited as part of issue #159 

Secondly this PR handles the upgrade of various upstream dependencies to now work with Numpy>= 2.0.


AlexanderWells-diamond:
Note there's now a pin for `numpy<2.0` in setup.py, which should be removed when this PR is worked on again.

AlexanderWells-diamond:
We still have a failure for Windows with Python 3.12, but that's waiting on a Numpy release after [this PR](https://github.com/numpy/numpy/issues/27224) was merged. We're going to merge these changes anyway, because once that is released everything _should_ just work...

Issue #162, opened 2024-06-27
Support for autosaving PV values


jsouter:
I have been working on an IOC to track the placement of samples on a carousel. It would be convenient if the readback values for the sample names at various positions would persist after an IOC restart in case of a power off event (or anything else prompting an IOC restart) so that they don't have to be manually scanned/written back in. I could rewrite as a normal C++ asyn IOC or implement my own autosave by writing to the filesystem but more official autosave support would be helpful.

olliesilvester:
MX would also find this useful. We have a use-case where we'd like to have a baton between GDA and Bluesky software, which is ran on a Python IOC. An autosave would be needed so it remembers who is in control after crashing / needing a restart.

coretl:
@jsouter I think my preferred approach would be a pure python interface like I07's. Can you think of a way to make it generic so it could be put into pythonSoftIOC?

jsouter:
I'll have a look at what's being done on i07 and see what I can come up with

jsouter:
Drafting this here: https://github.com/jsouter/pythonSoftIOC/tree/autosave
Currently you have to instantiate the Autosave class and pass to it a list of PVs you want to save, trying to inspect if there's a good way to use some builder magic to get a list of the records so that we could use .req files instead and look for PVs globally by name, though maybe explicitly passing the PV objects (RecordWrappers, technically) to Autosave is preferable. Right now the onus on the IOC developer to periodically call save in their main dispatcher loop, though it could be reworked so that there's always an autosave loop running in the background that periodically calls save if the save_period has elapsed and any of the PVs have changed.

Araneidae:
I suggest two things:
* Use an extra builder argument (maybe `autosave=True`) to automatically mark PVs for autosaving
* Run a background thread to periodically check for changes

It might be worth having a look at how [epics_device](https://github.com/dls-controls/epics_device) does things.  However, this is written in C and so may be a less accessible reference.

AlexanderWells-diamond:
I echo Michael's advice of using another flag during the build process to mark PVs for autosaving. Seems much cleaner than trying to maintain a separate list of PVs to be saved.

Although with the mention of `.req` files, this sounds like you have an alternative idea of how the interface should work. Is that file the one that is used by the EPICS autosave feature?

P.S. You can inspect the list of created records by calling `softioc.device_core.LookupRecordList()`

jsouter:
[...]

PR #163, opened 2024-07-03
Autosave support

jsouter:
Addressing https://github.com/dls-controls/pythonSoftIOC/issues/162
Adding in support for EPICS autosave style periodic saving of PV values and fields to a yaml-based backup file, from which initial values can be loaded on a restart.


softioc/asyncio_dispatcher.py
```diff
@@ -41,6 +42,13 @@ def __init__(self, loop=None, debug=False):
             raise ValueError("Provided asyncio event loop is not running")
         else:
             self.loop = loop
+        # set up autosave thread
+        autosaver = autosave.Autosave()
+        self.__autosave_worker = threading.Thread(
+            target=autosaver.loop,
+        )
+        self.__autosave_worker.daemon = True
+        self.__autosave_worker.start()
```
coretl, line 51: Probably want to have
```python
autosaver.stop()
self.__autosave_worker.join()
```
called in `self.__shutdown`

softioc/autosave.py
```diff
@@ -0,0 +1,131 @@
...
+                print(f"{name} is not a valid autosaved PV")
+                continue
+            pv.set(value)
+
+    def loop(self):
+        if not self._pvs:
+            return  # end thread if no PVs to save
+        while True:
+            time.sleep(self.save_period)
+            self.save()
```
coretl, line 131: E.g. if `self._stop_event = threading.Event()` in `__init__`
```suggestion
    def stop(self):
        self._stop_event.set()
        
    def loop(self):
        if not self._pvs:
            return  # end thread if no PVs to save
        while True:
            try:
                self._stop_event.wait(timeout=self.save_period)
            except TimeoutError:
                # No stop requested, we should save and continue
                self.save()
            else:
                # Stop requested
                return
```

softioc/cothread_dispatcher.py
```diff
@@ -20,6 +21,13 @@ def __init__(self, dispatcher = None):
             self.__dispatcher = dispatcher
 
         self.wait_for_quit = cothread.WaitForQuit
+        # set up autosave thread
+        autosaver = autosave.Autosave()
+        self.__autosave_worker = threading.Thread(
+            target=autosaver.loop,
+        )
+        self.__autosave_worker.daemon = True
+        self.__autosave_worker.start()
```
coretl, line 30: ditto

softioc/cothread_dispatcher.py
```diff
@@ -40,3 +47,7 @@ def wrapper():
             if completion:
                 completion(*completion_args)
         self.__dispatcher(wrapper)
+
+    def __shutdown(self):
```
coretl, line 51: There's no cleanup for the cothread dispatcher at the moment though, so this will never be called. Maybe put in a separate ticket for that though...

jsouter: I believe it does get called because of the atexit.register, I will test

jsouter:
Just tested with all the builder record types, looks like it's not happy with waveform record types (numpy array not serialisable to json), will either have to cast these to lists or try dumping to a yaml file, possibly with custom representers.

softioc/autosave.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def loop(self):
+        if not self._pvs:
+            return  # end thread if no PVs to save
+        while True:
+            self._stop_event.wait(timeout=self.save_period)
+            if self._stop_event.is_set():  # Stop requested
+                return
+            else:  # No stop requested, we should save and continue
+                self.save()
```
AlexanderWells-diamond, line 178: We need to consider how this should behave for long-running applications. Currently if there is any exception the entire thread will end up terminating. We probably don't want that, especially as it's possible that the exception could be from a temporary or one-time issue (e.g. network file mount blip).

We probably want at least a `try-except` clause here at this top level.

AlexanderWells-diamond (changes requested):
Looks good overall, but I've raised a few questions about what exactly the API could/should be in various comments. 

One more thing to think about is long-term reliability. If anything in `loop` has an Exception

There's several bits that need doing:
- Tests; at least a couple of system-level tests, then as much unit testing as you feel appropriate. Any test that starts an IOC will need to use the `multiprocessing` mechanism, as various parts of EPICS won't allow multiple IOC initializations in a single thread so a new one must be used for each IOC. A simple template is probably `test_record_wrapper_str` in `test_records.py`.
- Providing example code, either inside the existing `docs/examples/example_*.ioc.py` file(s) or in a separate new one
- Docs; Probably fine to make a new `how-to` document explaining how to configure and use autosave.
- A `CHANGELOG` entry - take a look at previous entries to see the format/syntax.

softioc/autosave.py
```diff
@@ -0,0 +1,178 @@
...
+
+SAV_SUFFIX = "softsav"
+SAVB_SUFFIX = "softsavB"
+
+def _ndarray_representer(dumper, array):
+    return dumper.represent_sequence(
+        "tag:yaml.org,2002:seq", array.tolist(), flow_style=True
+    )
+
+def configure(directory=None, save_period=None, device=None):
```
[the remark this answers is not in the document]
jsouter: Possibly, my feeling here was that this makes it closer to how the the ioc dev calls the builder module
`
from softioc import autosave, builder

builder.SetDeviceName("MY-DEVICE-NAME")
autosave.configure(directory="./autosave/MY-DEVICE-NAME")

`

softioc/autosave.py
```diff
@@ -0,0 +1,178 @@
...
+    else:
+        Autosave.directory = Path(directory)
+
+def set_autosave(pv, value=True):
+    if value:
+        Autosave.add_pv(pv)
+    else:
+        Autosave.remove_pv(pv)
+
+def load_req_file(file, override=False):
```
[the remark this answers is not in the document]
jsouter: Yeah, it's just a plaintext file that is a newline separated list of PV names that we want tracked by autosave

softioc/autosave.py
```diff
@@ -0,0 +1,178 @@
...
+    if directory is None and Autosave.directory is None:
+        raise RuntimeError(
+            "Autosave directory is not known, call "
+            "autosave.configure() with keyword argument "
+            "directory."
+        )
+    else:
+        Autosave.directory = Path(directory)
+
+def set_autosave(pv, value=True):
```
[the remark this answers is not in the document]
jsouter: My thoughts here was that the user could call set_autosave(pv, True) in a if __name__ in "__main__" block on a few PVs without changing the rest of the program where the records are defined before loading the database/calling IocInit(). I believe an error is raised if we attempt to change the pv's autosave attribute after builder.LoadDatabase() has been called - perhaps this is actually an argument for making it not visible then if it implies it can be called at runtime.

softioc/autosave.py
```diff
@@ -0,0 +1,178 @@
...
+            for pv_name in pv_names:
+                pv = LookupRecord(pv_name)
+                set_autosave(pv, True)
+        else:  # explicitly set autosave for False if pv not in file
+            for pv_name, pv in LookupRecordList():
+                set_autosave(pv, pv_name in pv_names)
+
+
+class Autosave:
+    _instance = None
```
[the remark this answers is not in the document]
jsouter: Whoops, yeah that's what the thinking was

softioc/autosave.py
```diff
@@ -0,0 +1,178 @@
...
+    _pvs = {}
+    save_period = 30.0
+    device_name = None
+    directory = None
+    enabled = True
```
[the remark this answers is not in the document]
jsouter: Not sure how I missed that but I intended autosave.configure(enabled=False) to allow disabling

softioc/cothread_dispatcher.py
```diff
@@ -13,13 +18,23 @@ def __init__(self, dispatcher = None):
...
         self.wait_for_quit = cothread.WaitForQuit
+        self.__atexit = atexit.register(self.__shutdown)
+
+        # set up autosave thread
+        self.__autosave = autosave.Autosave()
+        self.__autosave_worker = threading.Thread(
+            target=self.__autosave.loop,
+        )
+        self.__autosave_worker.daemon = True
+        self.__autosave_worker.start()
```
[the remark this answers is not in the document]
jsouter: Yep, I agree, I think I was trying to get around some circular dependency issues I've now resolved, seems like a more reasonable place, unless we feel that the implementation of the autosave thread should be more asyncio-y or cothread-y depending on our choice of dispatcher, or if we're happy to use a normal threading.Thread

softioc/cothread_dispatcher.py
coretl: You're right, I missed that...

Araneidae (changes requested):

jsouter:
I have added support for autosaving non-VAL fields, which is done by calling
builder.aOut("PV-NAME", autosave_fields=["SCAN", "PREC"]) for example. The autosave=True argument is used to determine if we add the VAL field to autosave, which I think is less ugly than requring the use to pass ["VAL"] to autosave_fields. (I note that this all needs to be added to the documentation when it's decided). I have noted however that In type records raise an exception when calling set_field, which gets called when loading from an autosave file. 
It could be possible to get around this by changing the setter to use the . accessor for fields, (pv.SCAN = "Passive") when loading, but we would have to make sure this load gets done before LoadDatabase() is called, which would be hard to enforce unless require the user to explicitly call something like a static autosave.load() before LoadDatabase() and softioc.iocInit().

AlexanderWells-diamond (changes requested):
The overall code and structure seems pretty good now. Unfortunately there's a number of bits causing problems, and a handful of tasks that still need doing.

There's a lot of CI test failures, some of which are due to the Pipfile.lock change, and some of them to do with the autosave code. 

For the Pipfile changes, there's a comment discussing whether they are needed or not.

For the other CI fails, I found some errors like this in a few places:
```
   ----------------------------- Captured stderr call -----------------------------
  2024-07-16T09:07:11.552312123 WARN pvxs.tcp.setup Server unable to bind port 5075, falling back to [::]:39597
  Process ForkServerProcess-5:
  Traceback (most recent call last):
    File "/opt/python/cp38-cp38/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
      self.run()
    File "/opt/python/cp38-cp38/lib/python3.8/multiprocessing/process.py", line 108, in run
      self._target(*self._args, **self._kwargs)
    File "/project/tests/test_record_values.py", line 384, in run_ioc
      builder.LoadDatabase()
    File "/tmp/tmp.ZezuRo2hXU/venv/lib/python3.8/site-packages/softioc/builder.py", line 303, in LoadDatabase
      autosave.load()
    File "/tmp/tmp.ZezuRo2hXU/venv/lib/python3.8/site-packages/softioc/autosave.py", line 70, in load
      Autosave._load()
    File "/tmp/tmp.ZezuRo2hXU/venv/lib/python3.8/site-packages/softioc/autosave.py", line 186, in _load
      cls._backup_sav_file()
    File "/tmp/tmp.ZezuRo2hXU/venv/lib/python3.8/site-packages/softioc/autosave.py", line 117, in _backup_sav_file
      sav_path = cls._get_current_sav_path()
    File "/tmp/tmp.ZezuRo2hXU/venv/lib/python3.8/site-packages/softioc/autosave.py", line 139, in _get_current_sav_path
      return cls.directory / f"{cls.device_name}.{SAV_SUFFIX}"
  TypeError: unsupported operand type(s) for /: 'NoneType' and 'str'
```

So it seems there are some cases where the autosave code is being initialized with invalid values.

The extra bits of work that still need doing:
- Documentation. `api.rst` needs extending to include the new keywords added to record initialization, and then we probably also need a new `how-to` doc to explain autosave and its mechanisms.
- A CHANGELOG entry - copy a previous one and change the info, as the syntax is quite picky!
- Possibly a new example python IOC using autosave (or modifications to an existing one)
- At least one full-system IOC test that actually starts an IOC from a canned autosave file and proves it loads the new values. Feel free to ask me when you get to this task as it is quite complicated to make it work.

softioc/autosave.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class _AutosavePV:
+    def __init__(self, pv, field=None):
+        if not field or field == "VAL":
+            self.get = pv.get
+            self.set = pv.set
+        else:
+            self.get = lambda: pv.get_field(field)
+            self.set = lambda val: setattr(pv, field, val)
```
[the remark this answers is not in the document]
jsouter: set_field can only be called successfully on Out type records by default, but we can set fields with the dot accessor before (and only before) the database is loaded on In and Out records. set_field can be called on In type records if we manually set DISP=False at PV init, so we could have some logic that temporarily overrides DISP until the load is completed, though this seemed hacky to me.

tests/test_autosave.py
```diff
@@ -0,0 +1,182 @@
...
+        assert pv_type in saved
+
+
+def test_can_save_fields(mocker, autosave_dir):
+    builder.SetDeviceName(DEVICE_NAME)
+    builder.aOut("SAVEVAL", autosave=True, autosave_fields=["DISA"])
+    builder.aOut("DONTSAVEVAL", autosave_fields=["SCAN"])
+    mocker.patch(
+        'softioc.device.ProcessDeviceSupportCore.get_field', return_value="0"
+    )
```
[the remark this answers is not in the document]
jsouter: Ah, no reason, I didn't see unittest in the setup configuration and forgot that it was included in Python by default, will port over.

docs/examples/example_autosave_ioc.py
```diff
@@ -0,0 +1,33 @@
+from softioc import autosave, builder, softioc
+import cothread
+
+# Set the record prefix
+builder.SetDeviceName("MY-DEVICE-PREFIX")
+
+# Create records, set some of them to autosave, also save some of their fields
+
+builder.aOut("AO", autosave=True)
+builder.aOut("AO2", autosave=True, autosave_name="EXTRA-AOUT")
```
Araneidae, line 10: What is the purpose of `autosave_name`?

docs/how-to/use-autosave-in-an-ioc.rst
```diff
@@ -0,0 +1,65 @@
...
+periodic saving to a backup file with the keyword arguments ``autosave``, ``autosave_fields``
+and ``autosave_name``. Records with ``autosave=True`` (``False`` by default) have their
+VAL fields backed up. Additional record fields in a list passed to ``autosave_fields`` will be backed
+up, note that this applies even when ``autosave`` is ``False``.
+
+The field values get written into a yaml-formatted file containing key-value pairs,
+by default the keys are the same as the record name passed as the first argument to the
+`builder.<RECORD_TYPE>()` call, excluding the device name specified in the `builder.SetDeviceName()`
+call, but this can be changed to a string name specified in the ``autosave_name`` keyword
+argument.
```
Araneidae, line 22: Why on earth would we want to do this?

jsouter: I personally wouldn't, by default we are stripping the device name from the autosave key to ensure consistency in the case where some PVs are instantiated before builder.SetDeviceName() and some are instantiated after. Users could manually set the autosave key to <device_name>.<pv_name> if they desired this, or any other naming scheme, though I suppose it's not necessary.

Araneidae: Unless there's a compelling use case (and this doesn't sound like it) I'd drop support for this option.

This interaction with `SetDeviceName()` sounds a bit troubling; can the name of a PV change after it has been created?  I thought the record name was fixed at the time of creation (the call to `aIn()` etc)?

jsouter: As far as I know there's no public way to change the record suffix, but the name that gets passed to the ProcessDeviceSupportCore (and subclass etc) init is record.name, see RecordWraper.__init__()

```
        record = builder(name, **fields)
        record.address = '@' + record.name
        self.__set('__builder', record)
        self.__set('__device',  device(record.name, **device_kargs))
```
record.name will either be <device_name>.<name> or just <name> if the prefix is not set, I think this name would otherwise become meaningless after LoadDatabase() is called. autosave_name could be removed if we passed the RecordWrapper.__init__ kwarg name directly instead of record.name in the device(record.name, device_kargs) call, though I haven't tested if anything could possibly break from doing that.

softioc/autosave.py
```diff
@@ -0,0 +1,240 @@
...
+                "directory"
+            )
+        if not self.directory.is_dir():
+            raise FileNotFoundError(
+                f"{self.directory} is not a valid autosave directory"
+            )
+        self._last_saved_time = datetime.now()
+
+    @classmethod
+    def _backup_sav_file(cls):
```
Araneidae, line 132: Why not use `__` prefix for local names?  This is the normal Python style for member names.

Araneidae (commented):
Looks good, have a few questions to look at.

softioc/device.py
```diff
@@ -53,6 +53,15 @@ class ProcessDeviceSupportCore(DeviceSupportCore, RecordLookup):
     # from record init or processing
     _epics_rc_ = EPICS_OK
 
+    # all record types can support autosave
+    def __init__(self, name, **kargs):
+        autosave_enabled = kargs.pop("autosave", False)
+        autosave_fields = kargs.pop("autosave_fields", [])
+        autosave_name = kargs.pop("autosave_name", name)
```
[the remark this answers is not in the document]
jsouter: This argument is only provided by softioc.pythonSoftIoc.RecordWraper.__init__ and is not publicly accessible to users - this is how we prevent the device prefix (which may or may not be set at time of PV creation), e.g. "name" could be set to "MY-DEVICE-PREFIX:MY-SIGNAL" or just "MY-SIGNAL", autosave_name will always be just "MY-SIGNAL", and this is the key that gets saved to the YAML file.

jsouter:
I will spend some time this afternoon looking back over original autosave and see if there's any features of that worth implementing. I'm not sure if there's any meaningful reason to try and implement things like the asVerify and the pass0/pass1 restoring of values before and after record initialization, but there are a few things like manual save restore that could be more relevant. I wonder if using the name autosave implies a sort of feature parity that we don't necessarily want

softioc/device.py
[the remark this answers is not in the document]
Araneidae: I see

AlexanderWells-diamond:
One last feature request: It may well be useful to provide a context manager, that can automatically do Autosave for all PVs declared within it. Possibly also supporting a specific list of fields, something like this:
```
with Autosave(autosave=True, autosave_fields=["EGU", "PREC"]):
  builder.aOut("MY_RECORD")
```
And this will autosave `MY_RECORD`'s VAL, EGU, and PREC fields.

jsouter:
> One last feature request: It may well be useful to provide a context manager, that can automatically do Autosave for all PVs declared within it. Possibly also supporting a specific list of fields, something like this:
> 
> ```
> with Autosave(autosave=True, autosave_fields=["EGU", "PREC"]):
>   builder.aOut("MY_RECORD")
> ```
> 
> And this will autosave `MY_RECORD`'s VAL, EGU, and PREC fields.

I'll push fixes to the above comments soon, I've got lots of things in branches at the moment! The most straightforward way I could think to do this was to define a context manager in builder.py like this 
```
_user_defaults = {}  # for use in DeviceFields context manager

class DeviceFields:
    def __init__(self, **fields):
        self._fields = fields
    
    def __enter__(self):
        global _user_defaults
        _user_defaults = self._fields
    
    def __exit__(self, A, B, C):
        global _user_defaults
        _user_defaults = {}

def _set_user_defaults(fields):
    fields.update(_user_defaults)
```
and have aOut etc call _set_user_defaults(fields). 

This is generic enough that it could be used for any EPICS field like EGU etc, would that be something we want? Or would it be better to rework the Autosave interface to work as a context manager?

```
with builder.DeviceFields(autosave=True, EGU="mm"):
    builder.aOut("MY-RECORD")
```

jsouter:
> I would prefer to see the Autosave class used as the context manager. As we don't have to run any code during the creation of each PV, what we can do is track how many PVs were created during the lifetime of the context manager and then add them in bulk to the Autosave list. Something like this (untested) code:
> 
> ```
> from softioc.device_core import LookupRecordList
> 
> class Autosave:
>     def __init__(self, **fields):
>         self._fields = fields
>     
>     def __enter__(self):
>        self.before = LookupRecordList()
>     
>     def __exit__(self, A, B, C):
>         after = LookupRecordList()
>         # Get diff of the keys in before and after dict, then add them to Autosave
> ```

Got it, I think this specific implementation would be problematic because of circular imports, but I can see if I can come up with something that achieves it. Of course could maybe avoid the import problem by importing within a class/function scope but could be a bit of a code smell

Araneidae:
> this specific implementation would be problematic because of circular imports

The simplest fix for that would be to replace
```python
from softioc.device_core import LookupRecordList
```
with
```python
from softioc import device_core
```
and invoke as `device_core.LookupRecordList`: Python handles circular imports by postponing populating a recursively imported module until the imports are all done, so you can import a module recursively so long as you don't look inside it too early!  (Maybe you already know all this...)

jsouter:
I realise I was mistaken about how the device name work, I was under the impression that the device name gets preprended to PVs that were instantiated before `builder.SetDeviceName()` when `builder.LoadDatabase()` gets called. Since this is apparently not the case, I will undo the change to strip the device prefix from the autosave key. This will make it much simpler to support the edge case where users change the device name in the middle of the program and define multiple records with the same name (which I have seen in a real softioc in use at Diamond). 

Considering the idea of multiple devices within an ioc, would we want to support the ability to backup differnet groups of PVs to multiple files, each with a different save period? I have drafted this with both sync and async methods, and it is a feature provided by C++ autosave.

AlexanderWells-diamond:
I'm inclined to keep it simple, at least for the first release. If anyone wants the extra feature they can raise an issue for it.

docs/examples/example_autosave_ioc.py
```diff
@@ -0,0 +1,34 @@
...
+# Set the record prefix
+builder.SetDeviceName("MY-DEVICE-PREFIX")
+
+# Create records, set some of them to autosave, also save some of their fields
+
+builder.aOut("AO", autosave=True)
+builder.aIn("AI", autosave_fields=["PREC", "EGU"])
+builder.boolIn("BO")
+builder.WaveformIn("WAVEFORMOUT", [0, 0, 0, 0], autosave=True)
+with autosave.Autosave(True, ["LOPR", "HOPR"]):
```
Araneidae, line 13: I wonder about the `True` argument here, it seems a bit redundant pretty well all of the time.  Howe about `Autosave` taking a `*args` list and defaulting to `"VAL"` if nothing is set?  Would that cover the normal bases and be sensible?  Am thinking you could code it something like:
```python
class Autosave:
    def __init__(self, *autosave_fields):
        if not autosave_fields: autosave_fields = ["VAL"]
        ...
```
Does mean people would have to write `"VAL"` sometimes which might be a surprise.

Araneidae (approved):
Looks good to me.  Would like to discuss the `Autosave` context handler default arguments, but think this is fine.

tests/test_autosave.py
```diff
@@ -0,0 +1,418 @@
...
+    ):
+        autosaver = autosave.Autosave()
+        autosaver._save()
+        with open(tmp_path / f"{DEVICE_NAME}.softsav", "r") as f:
+            saved = yaml.full_load(f)
+        assert "MANUAL" in saved
+        assert "MANUAL.EGU" in saved
+        assert "AUTOMATIC" in saved
+        assert "AUTOMATIC.PINI" in saved
+        assert "AUTOMATIC-OVERRIDDEN" in saved
```
AlexanderWells-diamond, line 234: Is it expected that the Context Manager's `True` would override the `autosave=False`? Or is there some other reason this item is in the saved list?

tests/test_autosave.py
```diff
@@ -0,0 +1,418 @@
...
+    autosave.Autosave.save_period = default_save_period
+    autosave.Autosave.device_name = default_device_name
+    autosave.Autosave.directory = default_directory
+    autosave.Autosave.enabled = default_enabled
+    autosave.Autosave.timestamped_backups = default_tb
+
+
+@pytest.fixture
+def existing_autosave_dir(tmp_path):
+    state = {
```
AlexanderWells-diamond, line 36: As this data is used in several tests, I'd be tempted to split it out into its own top level dictionary or a fixture, to make it obvious that these values will be references in many tests. Not required, just a nice to have.

AlexanderWells-diamond (changes requested):
Overall this is good, very close to done. I spotted a couple of potential bugs and there's a few docs changes that I'd like to see before this is finished.

docs/reference/api.rst
```diff
@@ -260,6 +329,31 @@ and stderr streams, is sent directly to the terminal.
     .. seealso::
         `SetBlocking` for configuring a global default blocking value
 
+    .. _autosave_arg:
+
+    `autosave`
```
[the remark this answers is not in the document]
jsouter: Ah okay, will try and fix. Using ".. _autosave" here creates warnings on a build because of the ambiguity between references to the autosave arg here and the softioc.autosave module in the docs.

softioc/autosave.py
```diff
@@ -0,0 +1,274 @@
...
+    Autosave.save_period = save_period
+    Autosave.enabled = enabled
+    Autosave.device_name = name
+
+
+def start_autosave_thread():
+    worker = threading.Thread(
+        target=Autosave._loop,
+    )
+    worker.daemon = True
```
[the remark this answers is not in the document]
AlexanderWells-diamond: After a chat with @Araneidae we think it may not be necessary to use a daemon thread. You already have a clean shutdown mechanism when using the Event, so by removing this line we should avoid the potential for file corruption on shutdown.

docs/how-to/use-autosave-in-an-ioc.rst
```diff
@@ -0,0 +1,73 @@
...
+
+    MY-DEVICE-PREFIX:AI.EGU: ''
+    MY-DEVICE-PREFIX:AI.PREC: '0'
+    MY-DEVICE-PREFIX:AO: 0.0
+    MY-DEVICE-PREFIX:AUTOMATIC-AO: 0.0
+    MY-DEVICE-PREFIX:AUTOMATIC-AO.EGU: ''
+    MY-DEVICE-PREFIX:AUTOMATIC-AO.HOPR: '0'
+    MY-DEVICE-PREFIX:AUTOMATIC-AO.LOPR: '0'
+    MY-DEVICE-PREFIX:MINUTESRUN: 1
+    MY-DEVICE-PREFIX:WAVEFORMOUT: [0, 0, 0, 0]    AI.EGU: ''
```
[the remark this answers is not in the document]
jsouter: Gah! I think probably an accidental middle mouse click

tests/test_autosave.py
jsouter: That's how it works currently yes, I suppose we could find a way to use some True/False/None logic if we wanted an explicit calling of autosave=False to override the context manager?

AlexanderWells-diamond: I see. My view would be that setting `autosave=False` on the record creation function should override everything else - it's the principle of least surprise to take its value over any other.

However, I can see that may be difficult to actually achieve. I can also see that it seems very unlikely anyone will ever actually try and do that. If there's an easy way to at least print/raise an error in this case then do that, otherwise it is a rather confusing behaviour. @Araneidae, any thoughts?

jsouter: Can roll back if required but just pushed something to address this: we call add_pv_to_autosave on all PVs at init now, and class variables are set in the context manager which are used to override the the arguments "autosave" and "autosave_fields" if not specified, so an explicit autosave=False will get applied. It also means we don't have to use LookupRecordList any more.

But I did realise that this newer implementation, and the LookupRecordList implementation would not be thread safe, in the sense that if a separate thread is creating PVs while another thread is creating PVs inside a context manager, we may accidentally track fields that are not specified. I'm not sure if anyone would write their IOC that way, but may be something worth trying to guard against.

jsouter:
I believe I have managed to make the context managers thread safe by using a singleton style class that inherits from threading.local. I have moved some of the context manager logic into a new _AutosaveContext class, but if this seems a bit too convoluted see the approach in the commit https://github.com/DiamondLightSource/pythonSoftIOC/pull/163/commits/2da2fbf823d5651500164954effafe6c287d1e64 which applies all the singleton/threading logic to Autosave.

jsouter:
If we have a non-default list of autosave_fields for a PV inside an Autosave context manager, do we want to combine the lists or give priority to just the specified fields in the PV init? I suppose we still need to discuss if we want to track VAL inside autosave_fields instead of using autosave as its own kwarg.

Araneidae:
> Regarding tracking VAL, it's an awkward one due to the assumption that a record name without `.VAL` is actually referring to the VAL field. If that's also how C++ Autosave works I'm inclined to leave it as it is - having `autosave=True` is what I expect most users will want, and is shorter than writing `autosave_fields=["VAL"]`.

A hacky workaround that occurs to me is this:
* Get rid of `autosave_fields`
* Make `autosave` behave mostly like `autosave_fields` does now but...
* Allow `autosave` to be set to one of three types of value with the following interpretations:

  * A list (of strings) means autosafe all the listed fields, as for `autosave_fields` now
  * A value of `True` means the same as `['VAL']`
  * A single string `s` means the same as `[s]` (this one may be a hack too far)

AlexanderWells-diamond:
I think I'm ok with @Araneidae 's suggestion of using a single variable as either a bool, a list, or none. It keeps the shorthand for enabling saving for VAL, while also allowing flexibility for more fields. Only potential downside is that you would have to specify VAL in the list e.g. `autosave=["VAL", "EGU"]` - a minor concern but one that would be worth ensuring is highlighted in the docs.

OCopping:
The builds are failing due to `actions/upload-artifact@v2` being deprecated (see: [here](https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/))
To fix this, bump to either `v3` or `v4`.
However if bumping to `v4`, you will need to make the artifact name unique for each build, for example in https://github.com/epics-base/epicscorelibs/pull/29

jsouter:
> > Regarding tracking VAL, it's an awkward one due to the assumption that a record name without `.VAL` is actually referring to the VAL field. If that's also how C++ Autosave works I'm inclined to leave it as it is - having `autosave=True` is what I expect most users will want, and is shorter than writing `autosave_fields=["VAL"]`.
> 
> A hacky workaround that occurs to me is this:
> 
> * Get rid of `autosave_fields`
> * Make `autosave` behave mostly like `autosave_fields` does now but...
> * Allow `autosave` to be set to one of three types of value with the following interpretations:
>   
>   * A list (of strings) means autosafe all the listed fields, as for `autosave_fields` now
>   * A value of `True` means the same as `['VAL']`
>   * A single string `s` means the same as `[s]` (this one may be a hack too far)

Have implemented all these, can remove support for passing a single string if we think it's too hacky.

tests/test_autosave.py
```diff
@@ -0,0 +1,492 @@
...
+    pv_thread_before_cm = threading.Thread(
+        target=create_pv_in_thread, args=["PV-FROM-THREAD-BEFORE", 1])
+    pv_thread_in_cm = threading.Thread(
+        target=create_pv_in_thread, args=["PV-FROM-THREAD-DURING", 0])
+    pv_thread_before_cm.start()
+    with autosave.Autosave(["VAL", "EGU"]):
+        builder.aOut("PV-FROM-CM")
+        pv_thread_in_cm.start()
+        pv_thread_in_cm.join()
+    pv_thread_before_cm.join()
```
AlexanderWells-diamond, line 443: On initial impression this test looks pretty flaky - relying on a `time.sleep` is not stable, especially across different platforms which have different threading behaviours. I've observed the MacOS GH runners taking significantly longer to spin up threads than other platforms, for example.

A better way would be to create a couple of `threading.Event` objects in the main level, and have the child threads wait on them until the main thread sets them. That will ensure much more consistency.

tests/test_autosave.py
```diff
@@ -0,0 +1,492 @@
...
+    builder.aOut("FIELDS", autosave=["LOPR", "HOPR"])
+    builder.aOut("FALSE", autosave=False)
+    builder.aOut("SINGLE-FIELD", autosave="EGU")
+    builder.aOut("AUTO-VAL", autosave="VAL")
+    assert set(autosave.Autosave._pvs) == {
+        "TRUE", "FIELDS.LOPR", "FIELDS.HOPR", "SINGLE-FIELD.EGU", "AUTO-VAL"}
+    autosave.Autosave._pvs = {}
+    builder.ClearRecords()
+    with autosave.Autosave():  # True by default
+        builder.aOut("AUTO-TRUE", autosave=False)
```
AlexanderWells-diamond, line 471: Something seems wrong here; specifying `autosave=False` is ignored (as it looks like this PV does appear in the saved PV list)? That's a different behaviour to below (lines 486-487) where specifying `autosave=True` does override the CM's False flag.

AlexanderWells-diamond (changes requested):
Final few things to clear up! Nearly there!

tests/test_autosave.py
jsouter: Okay, setting autosave=False explicitly disables autosave for the pv inside a context manager now.

AlexanderWells-diamond (approved):
Looks good to me!

jsouter:
Great! I don't seem to have permissions to merge but happy to have it squashed with a message like "add support for saving and restoring pv fields with autosave"

PR #164, opened 2024-07-11
Ensure returned numpy arrays are not writeable

AlexanderWells-diamond:
As we return a reference to the numpy array, rather than a copy, we want to ensure that the value is not modified under our noses. 

Now when you try and modify a returned array there will be a ValueError raised.

See documentation here:
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html

### DIFF
--- a/softioc/device.py
+++ b/softioc/device.py
@@ -467,7 +467,12 @@ class WaveformBase(ProcessDeviceSupportCore):
         # common class of bug, at the cost of duplicated code and data, here we
         # ensure a copy is taken of the value.
         assert len(value) <= self._nelm, 'Value too long for waveform'
-        return numpy.copy(value)
+
+        value = numpy.copy(value)
+        # As we return a reference to the numpy array, ensure it cannot be
+        # modified under our noses
+        value.flags.writeable = False
+        return value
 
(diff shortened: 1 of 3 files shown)


softioc/device.py
```diff
@@ -473,6 +473,9 @@ def _epics_to_value(self, value):
         if self._dtype.char == 'S':
             return [_string_at(s, 40) for s in value]
         else:
+            # As we return a reference to the numpy array, ensure it cannot be
+            # modified under our noses
+            value.flags.writeable = False
```
Araneidae, line 478: Probably better to do this on the _incoming_value, so do this a few lines above here https://github.com/dls-controls/pythonSoftIOC/blob/8e555c2b95bd3d0c3f8c05b8b9fbd2ea4a1db27e/softioc/device.py#L470 before returning the copy.

Araneidae (approved):
This is technically a breaking change, but not I think worth a major version bump.  It would be nice to check if this does break any naughty code ... topup seems a good candidate to check!

softioc/device.py
AlexanderWells-diamond: Done in  8c4b516

PR #173, opened 2024-09-04
Fix broken link for `isues`.

gustavosr8:


Araneidae:
Hi @gustavosr8, this PR made me take a closer look at the links and I found a number of other broken links which I have now fixed as well.

PR #186, opened 2025-03-27
Allow starting IOC without PVA

### LINKED ISSUE #185: Starting an IOC without QSRV
Could we have a way to start a soft IOC up, without QSRV? I believe we'd need to remove the following lines from `__init__.py`:

```python
dbLoadDatabase('pvxsIoc.dbd', pvxslibs.path.dbd_path, None)
ctypes.CDLL(find_dso('pvxslibs.lib.pvxsIoc'), ctypes.RTLD_GLOBAL)
os.environ.setdefault('PVXS_QSRV_ENABLE', 'YES')
```
to achieve this.

AlexanderWells-diamond:
There seems to be no need to load the DBD and DSO early, so doing it just before iocInit allows a programmatic switch to enable/disable it.

Fixes #185

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools", "wheel", "setuptools_dso>=2.1", "epicscorelibs>=7.0.7.99.1.2a1"]
+requires = ["setuptools", "wheel", "setuptools_dso>=2.1", "epicscorelibs>=7.0.7.99.1.1a3"]
 build-backend = "setuptools.build_meta:__legacy__"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools", "wheel", "setuptools_dso>=2.1", "epicscorelibs>=7.0.7.99.1.1a3"]
+requires = ["setuptools", "wheel", "setuptools_dso>=2.1", "epicscorelibs>=7.0.7.99.1.2a1"]
 build-backend = "setuptools.build_meta:__legacy__"
--- a/softioc/softioc.py
+++ b/softioc/softioc.py
@@ -6,7 +6,7 @@ from ctypes import *
 from tempfile import NamedTemporaryFile
 
 import pvxslibs.path
-from epicscorelibs.ioc import registerRecordDeviceDriver, pdbbase
+from epicscorelibs.ioc import pdbbase
 from setuptools_dso.runtime import find_dso
 
 from . import autosave, imports, device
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools", "wheel", "setuptools_dso>=2.1", "epicscorelibs>=7.0.7.99.1.1a3"]
+requires = ["setuptools", "wheel", "setuptools_dso>=2.1", "epicscorelibs>=7.0.7.99.1.2a1"]
```
Araneidae, line 2: Eww, seems a bit arbitrary.  This pinning looks all a bit uncomfortable to be, I suppose there's a purpose for it?

softioc/softioc.py
```diff
@@ -33,6 +39,14 @@ def iocInit(dispatcher=None):
         dispatcher = cothread_dispatcher.CothreadDispatcher()
     # Set the dispatcher for record processing callbacks
     device.dispatcher = dispatcher
+
+    if enable_pva:
+        dbLoadDatabase('pvxsIoc.dbd', pvxslibs.path.dbd_path, None)
+        ctypes.CDLL(find_dso('pvxslibs.lib.pvxsIoc'), ctypes.RTLD_GLOBAL)
+
+        if registerRecordDeviceDriver(pdbbase):
+            raise RuntimeError('Error registering')
```
Araneidae, line 48: Very un-Pythonic error handling here!

I suggest moving `registerRecordDeviceDriver` to [imports.py](https://github.com/DiamondLightSource/pythonSoftIOC/blob/9eb29d037d5ec700bcb2616b8b3bfea16b6b64dd/softioc/imports.py) and adding `.errcheck = expect_success` to its definition.

Araneidae (commented):
Short sweet and to the point, but I suggest improving the error handling.

softioc/softioc.py
```diff
@@ -33,6 +39,13 @@ def iocInit(dispatcher=None):
         dispatcher = cothread_dispatcher.CothreadDispatcher()
     # Set the dispatcher for record processing callbacks
     device.dispatcher = dispatcher
+
+    if enable_pva:
+        dbLoadDatabase('pvxsIoc.dbd', pvxslibs.path.dbd_path, None)
```
coretl, line 44: Is there any reason we shouldn't defer all dbd and lib loading until iocInit?

softioc/softioc.py
AlexanderWells-diamond: Done in 026ed69c20d9a92c729e022fbfbc999aef546d08

Araneidae (approved):

softioc/softioc.py
AlexanderWells-diamond: With this PR there is no longer any lib loading anywhere except in `iocInit()`.

You must load the DBD files before creating records, otherwise epicsdbbuilder / EPICS itself doesn't know what to do with the record entries. So I believe we must do that as early as possible.

For research I tried moving the DBD loading into this function, and it results in a segmentation fault as soon as you try and create a record (presumably EPICS is attempting to call some kind of validation function related to the DBD and doesn't have proper null checks somewhere)

PR #187, opened 2025-03-28
Remove unnecessary __super and __init__class__

### LINKED ISSUE #167: Remove __super attribute
pythonSoftIOC has [code which adds __super](https://github.com/DiamondLightSource/pythonSoftIOC/blob/6f8920baacceca1c6bfd547e45eb04920a3d0dec/softioc/device_core.py#L27) as an attribute to classes, to avoid the old semantics in Python2 of having to specify `super(A, self)` to access a `super` method. In Python3 all we need is `super()`. It is likely we can remove this attribute setting and make use of the new, cleaner, syntax.

AlexanderWells-diamond:
Python3 fully supports use of super() without additional required parameters, unlike Python2.

`__init__class__` and the associated metaclass logic are no longer required as `__init__subclass__` does exactly what we want; running once per subclass created.

Fixes #167

### DIFF
--- a/softioc/device_core.py
+++ b/softioc/device_core.py
@@ -5,7 +5,7 @@ from .fields import RecordFactory
 from ctypes import *
 
 
-class DeviceCommon():
+class DeviceCommon:
     # By requiring that DeviceCommon be a common base class for the entire
     # Python device hierarchy, we can use this __init__ to test for unused
     # keyword arguments.
--- a/softioc/device_core.py
+++ b/softioc/device_core.py
@@ -1,46 +1,11 @@
 from __future__ import print_function
 
-import sys
 from . import imports
 from .fields import RecordFactory
 from ctypes import *
 
 
-# Black magic lifted from six.py (http://pypi.python.org/pypi/six/) to replace
-# use of __metaclass__ for metaclass definition
-def with_metaclass(meta, *bases):
-    class metaclass(meta):
-        def __new__(cls, name, this_bases, d):
-            return meta(name, bases, d)
-    return type.__new__(metaclass, 'temporary_class', (), {})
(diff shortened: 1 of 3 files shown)


softioc/device_core.py
```diff
@@ -1,46 +1,11 @@
...
-        cls.__init_class__()
-
-class DeviceCommon(with_metaclass(InitClass)):
-    '''Adds support for an __init_class__ method called when the class or any
-    of its subclasses is constructed.  Also adds auto-super functionality
-    (see iocbuilder.support.autosuper).'''
-
-    def __init_class__(cls): pass
-
+class DeviceCommon():
```
Araneidae, line 8: Don't think the brackets `()` are needed here, are they?

softioc/device_core.py
```diff
@@ -83,20 +48,16 @@ class DeviceSupportCore(DeviceCommon):
...
         if not hasattr(cls, '_record_type_'):
             # If no record type has been specified then we're a base class
             # with nothing to do.
             return
 
-        # Create an empty device directory.
-        # (Potentially this belongs in a mix-in for resolving the linkage
-        # between record and instances, but for now this is hard-wired here.)
-        cls.__device_directory = {}
-
```
Araneidae, line 99: Interesting: I can find no trace of this symbol anywhere else (even using `git log -S`)!  I guess this was an idea that either never happened or has been erased.

Araneidae (approved):
Nice and simple.

softioc/device_core.py
AlexanderWells-diamond: Removed in 9b243807d473fe1daa1b067fc310eb7e870eb09d

